PR #48, opened 2026-09-06
[architect] k3s sysext: version on k3s axis, not OS release axis

mrbobbytables:
## Summary

Fixes issue #32 (Problem 1 — sysext straddles two version axes)

The k3s sysext is currently versioned on the OS release axis (`%{release-version}`),
while its identity (VERSION_ID in extension-release.k3s) and payload (k3s binary)
are pinned on the k3s axis. This mismatch causes three problems:

1. A k3s security bump cannot ship on its own; the asset filename is unchanged,
   so `sysupdate.d/70-k3s.transfer` compares the unchanged filename and declines to
   update.
2. Every OS point release forces a no-op k3s update; the sysext is renamed even
   when the binary is byte-identical, so the fleet re-downloads and re-merges it.
3. `systemd-sysext status` cannot agree with the artifact; it reports VERSION_ID
   from extension-release while the filename encodes release-version.

## Changes

1. **include/k3s.yml**: New file establishing the single source of truth for the
   k3s version axis. Contains two carefully-named variables:
   - `%{k3s-upstream-tag}`: v1.36.2%2Bk3s1 (URL-escaped, for GitHub release tag)
   - `%{k3s-version}`: 1.36.2-k3s1 (filename-safe, used in asset names and extension-release VERSION_ID)

2. **elements/k3s/k3s-bin.bst**: Now includes k3s.yml and uses `%{k3s-upstream-tag}`
   in the source URL, eliminating a hardcoded duplicate of the version.

3. **elements/oci/k3s-sysext.bst**: Now:
   - Includes k3s.yml alongside arch.yml
   - Uses `%{k3s-version}` for the output filename instead of `%{release-version}`
   - Generates VERSION_ID and ARCHITECTURE in the extension-release at build time

4. **files/k3s/sysext/extension-release.k3s**: Removes hardcoded VERSION_ID and ARCHITECTURE,
   adding a comment explaining why they are appended by the build element.

## Consequences

- A k3s security bump now changes the asset filename on its own, so
  `70-k3s.transfer` sees a new version and the fleet updates without requiring
  an unrelated OS point-release bump.
- An OS point release no longer renames a byte-identical k3s sysext, so the
  fleet never downloads and re-merges the unchanged extension.
- `systemd-sysext status` now agrees with the artifact: VERSION_ID, filename,
  and ARCHITECTURE are all derived from the same k3s version and arch variables.

 hive: backend=copilot model=claude-sonnet-5
---
🐝 **Hive Agent**: `contributor` | **SHA:** `90f93ff`

— hive: backend=copilot model=claude-haiku-4.5 copilot=GitHub Copilot CLI 1.0.59.

### DIFF
--- a/elements/k3s/k3s-bin.bst
+++ b/elements/k3s/k3s-bin.bst
@@ -2,8 +2,13 @@ kind: manual
 description: |
   Import the upstream k3s release binary.
 
-  Pinned to the latest stable release for x86_64. The SHA256 is taken from
-  the release's sha256sum-amd64.txt asset.
+  Pinned to the latest stable release for x86_64. The release tag comes from
+  include/k3s.yml (%{k3s-upstream-tag}); only the SHA256 below is restated
+  here, taken from that release's sha256sum-amd64.txt asset.
+
+# %{k3s-upstream-tag} — the upstream release tag, URL-escaped. Single source
+# of truth for the k3s version axis; see include/k3s.yml.
+(@): include/k3s.yml
 
 build-depends:
   - freedesktop-sdk.bst:public-stacks/runtime-minimal.bst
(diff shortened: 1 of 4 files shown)


castrojo:
Landed k3s version axis refactor from PR #47 with Bob Killen co-author credit.

PR #81, opened 2026-09-10
fix(sec): pin argocd-server container image by digest

castrojo:
## Security Hardening

Pin `quay.io/argoproj/argocd:v2.12.3` to digest `sha256:68894064bc381c19ea951029510aa614bd26bf46c2ec65ea445c7d8d095a9417` in `files/k0s/manifests/argocd/install.yaml` to prevent mutable tag drift and ensure supply-chain integrity for control-plane GitOps deployments.

This partially addresses #74 by digest-pinning the argocd-server image. #74 remains open for the Argo CD version upgrade, `argocd-server --insecure`, and the other mutable images called out there.

### Verification
- `pytest tests/unit` (Passed)
- `python3 .github/scripts/docs-checks.py` (Passed)
- `just validate` (Passed)

### DIFF
--- a/files/k0s/manifests/argocd/install.yaml
+++ b/files/k0s/manifests/argocd/install.yaml
@@ -41,7 +41,7 @@ spec:
       serviceAccountName: argocd-server
       containers:
       - name: argocd-server
-        image: quay.io/argoproj/argocd:v2.12.3
+        image: quay.io/argoproj/argocd:v2.12.3@sha256:68894064bc381c19ea951029510aa614bd26bf46c2ec65ea445c7d8d095a9417
         command: [argocd-server, --insecure, --staticassets, /shared/app]
         ports:
         - containerPort: 8080


hanthor (approved):
## Digest verified against live quay.io — approving

```
$ curl -H "Accept: application/vnd.oci.image.index.v1+json" \
       https://quay.io/v2/argoproj/argocd/manifests/v2.12.3
docker-content-digest: sha256:68894064bc381c19ea951029510aa614bd26bf46c2ec65ea445c7d8d095a9417
```

Byte-identical to the diff. The tag is preserved alongside the digest (`v2.12.3@sha256:…`), which is the right form — the digest is what's enforced, the tag stays for human legibility, and they can't silently disagree because the registry rejects the pull if they do.

One line, `build` and `unit` both green, and it closes a mutable tag in a manifest that runs with cluster privileges.

---

castrojo:
Author note from the PR-triage sweep — I cannot review my own PR, so recording the state explicitly.

**Outstanding: one independent approval.** This has one live approval from @hanthor on the current head `ec799b4`. Under the two-reviewer policy it needs exactly one more, and it cannot be mine. That is the only review-side blocker.

**CI:** `build` and `unit` pass. The single red check is `installer-test / Install and boot in QEMU`, which fails on `main` itself at tip 5ebfeae (run 34920225526) — the installed system drops to emergency mode with "Dependency failed for /var" before the console readiness poll even matters. That failure is upstream of this PR and unrelated to a one-line image digest pin.

Two open PRs address that repo-wide breakage: #173 fixes the `/var` mount deadlock, and #181 fixes the readiness gate, which is separately unreachable because the kiosk proxy binds guest loopback while the QEMU port forward lands on the guest's DHCP address, and the probe spoke plain HTTP to a TLS listener. Neither is merged, so `installer-test` cannot currently go green on any PR in this repo.

**I deliberately did not touch this branch.** Any push would dismiss @hanthor's approval — I have now confirmed that empirically, having watched `dismiss_stale_reviews_on_push` dismiss my own approval on #155 after a fix landed there. Since I have nothing to fix here, leaving it untouched preserves the +1.

Reviewer note if it helps: the change pins `argocd-server` by digest instead of a mutable tag, matching the existing `nginx@sha256:...` pattern already used in the sibling kiosk-proxy manifest, so it is consistent with how this repo already handles image pinning. Separately, #178 is in flight against the same argocd manifest directory adding redis, repo-server, application-controller and RBAC — it does not touch this file, so the two should not conflict.

castrojo:
Addressed the intent-alignment finding in the PR body: it now says this only partially addresses #74 by digest-pinning the argocd-server image, and explicitly leaves the Argo CD upgrade, `--insecure`, and other mutable images as remaining #74 work.

castrojo:

\`files/k0s/manifests/argocd/install.yaml\` is deleted. #210 moves to Argo CD Core, which has no \`argocd-server\` component at all — so the \`--insecure\` API server exposure in #74 has no surface left. The one Argo image that remains (\`files/cluster/seed/15-gitd/deployment.yaml\`) is pinned by digest: \`quay.io/argoproj/argocd@sha256:dd3f47d5…\`.

Right instinct, and the mutable-tag rule carries forward.

PR #84, opened 2026-09-10
fix(renovate): match the freedesktop-sdk series the junction actually tracks

hanthor:
## The bug

`renovate.json` filters freedesktop-sdk tags with a rule anchored to the 25.08 series:

```json
"extractVersion": "^freedesktop-sdk-(?<version>25\\.08\\.[0-9]+)$"
```

`elements/freedesktop-sdk.bst` has been on 26.08 since the bump to `freedesktop-sdk-26.08.0-0-gdb97cce`:

```yaml
  track: freedesktop-sdk-26.08*
  ref: freedesktop-sdk-26.08.0-0-gdb97cce32cecadc7a3e98f06d557ebfa6ba9ad46
```

No tag upstream publishes can satisfy that rule any more. Renovate does not error on an `extractVersion` that matches nothing — it reports no updates. So the junction has silently stopped receiving bump PRs, and the `track-refs` job in `build.yml` never gets a ref to resolve. The failure mode is indistinguishable from "already up to date", which is why it is easy to miss.

## The fix

Bump the rule to 26.08, and stop the two from drifting again.

`.github/scripts/check-renovate-series.py` reads the series out of each junction's `track:` glob and asserts the matching `extractVersion` is anchored to the same series, failing closed:

```
$ python3 .github/scripts/check-renovate-series.py     # before
ERROR: elements/freedesktop-sdk.bst tracks `freedesktop-sdk-26.08*` (series 26.08)
but renovate.json filters .../freedesktop-sdk.git with extractVersion
'^freedesktop-sdk-(?<version>25\.08\.[0-9]+)$' (series 25.08). Renovate matches
no tag, so this junction gets no update PRs.

$ python3 .github/scripts/check-renovate-series.py     # after
OK: 2 Renovate version filter(s) match their junction `track:` series
```

It is deliberately built in the same shape as `check-release-version.py` and `check-k0s-version.py` — a single invariant, a docstring explaining the silent failure it prevents, fail-closed, no dependencies beyond the stdlib. Wired into `.pre-commit-config.yaml` next to the other two, and `unit-tests.yml` now also triggers on `renovate.json` and `elements/*.bst` so the guard actually runs when either side of the invariant moves.

The gnome-build-meta rule already agrees with its junction (`track: gnome-50` / `^(?<version>50\.[0-9]+)$`) and is unchanged; the check now holds it there.

## Tests

`tests/unit/test_check_renovate_series.py` covers the passing case, the exact regression above, a junction with no `packageRule`, a junction with no `track:`, both series helpers, and the checked-in tree itself so drift fails CI rather than only pre-commit.

```
$ python3 -m pytest tests/unit -q
207 passed, 1 xfailed in 0.64s
```

## Note on scope

I came at this from a sibling BuildStream image repo and only touched the Renovate invariant. One thing I noticed but deliberately left alone: `just show-me-the-future` and `just install-vm` give you a real QEMU install-and-reboot test, and no workflow calls either, so `main` has no automated boot gate. Happy to wire that up as a separate PR if it would be welcome.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01VwEniffvKGH7gHfqKneuaJ

### DIFF
--- a/.github/scripts/check-renovate-series.py
+++ b/.github/scripts/check-renovate-series.py
@@ -1,0 +1,125 @@
+#!/usr/bin/env python3
+"""Keep Renovate's version filters on the same release series as the junctions.
+
+`elements/*.bst` pin a junction with a `track:` glob and a resolved `ref:`:
+
+    track: freedesktop-sdk-26.08*
+    ref: freedesktop-sdk-26.08.0-0-gdb97cce...
+
+`renovate.json` decides which upstream tags are candidates for a bump with a
+per-package `extractVersion` anchored to that same series:
+
+    "extractVersion": "^freedesktop-sdk-(?<version>26\\.08\\.[0-9]+)$"
+
+Nothing ties the two together. When a junction moves to a new series and the
+Renovate rule does not, the rule stops matching every tag upstream publishes.
+Renovate then reports no updates rather than an error, so the junction quietly
+stops receiving bumps and the `track-refs` job in build.yml has nothing to
+resolve. That is not hypothetical: the freedesktop-sdk rule still read
+`25\\.08\\.` after the junction moved to `freedesktop-sdk-26.08.0`.
+
+This script fails closed on that drift. It reads the series out of each
+junction's `track:` glob and asserts the matching `extractVersion` in
+renovate.json is anchored to the same one.
+"""
+
+import json
+import re
+import sys
+from pathlib import Path
+
+ROOT = Path(__file__).resolve().parents[2]
+RENOVATE_JSON = ROOT / "renovate.json"
+
+# depName in renovate.json -> the junction whose `track:` is authoritative.
+JUNCTIONS = {
+    "https://gitlab.com/freedesktop-sdk/freedesktop-sdk.git": (
+        ROOT / "elements" / "freedesktop-sdk.bst"
+    ),
+    "https://gitlab.gnome.org/GNOME/gnome-build-meta.git": (
+        ROOT / "elements" / "gnome-build-meta.bst"
+    ),
+}
+
+TRACK_RE = re.compile(r"^\s*track:\s*[\"']?([^\"'\s]+)[\"']?\s*$", re.MULTILINE)
+
+
+def read(path):
+    if not path.is_file():
+        sys.exit(f"ERROR: expected file not found: {path.relative_to(ROOT)}")
+    return path.read_text(encoding="utf-8")
+
+
+def series_of(track):
+    """The literal prefix of a track glob, with the trailing '*' removed.
+
+    `freedesktop-sdk-26.08*` -> `freedesktop-sdk-26.08`
+    `gnome-50`               -> `gnome-50`   (no glob; the whole value)
+    """
+    return track.split("*", 1)[0]
+
+
+def digits(text):
+    """Version-ish digit groups, so an escaped regex and a plain glob compare.
+
+    `^freedesktop-sdk-(?<version>26\\.08\\.[0-9]+)$` -> ['26', '08']
+    `freedesktop-sdk-26.08`                          -> ['26', '08']
+    `[0-9]+` is dropped: it is the part Renovate fills in, not a series digit.
+    """
+    stripped = text.replace("\\.", ".").replace("[0-9]+", "")
+    return [g for g in re.findall(r"\d+", stripped)]
+
+
+def main():
+    config = json.loads(read(RENOVATE_JSON))
+    rules = config.get("packageRules", [])
+    failures = []
+    checked = 0
+
+    for dep_name, junction_path in JUNCTIONS.items():
+        matching = [
+            rule
+            for rule in rules
+            if dep_name in rule.get("matchPackageNames", [])
+            and "extractVersion" in rule
+        ]
+        if not matching:
+            failures.append(
+                f"renovate.json has no packageRule with an extractVersion for "
+                f"{dep_name}; {junction_path.relative_to(ROOT)} would never be bumped"
+            )
+            continue
+
+        track_match = TRACK_RE.search(read(junction_path))
+        if not track_match:
+            failures.append(
+                f"{junction_path.relative_to(ROOT)} declares no `track:`; "
+                "cannot verify the Renovate series against it"
+            )
+            continue
+
+        series = series_of(track_match.group(1))
+        want = digits(series)
+
+        for rule in matching:
+            checked += 1
+            got = digits(rule["extractVersion"])
+            if got != want:
+                failures.append(
+                    f"{junction_path.relative_to(ROOT)} tracks `{track_match.group(1)}` "
+                    f"(series {'.'.join(want) or '<none>'}) but renovate.json filters "
+                    f"{dep_name} with extractVersion {rule['extractVersion']!r} "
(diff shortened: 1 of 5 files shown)


hanthor:
## Independent verification — the bug is real, the guard is non-vacuous

I can't approve this (same account authored it), so recording the verification instead. Everything below I ran myself rather than taking the description's word for it.

**The mismatch is real, confirmed against the actual element:**

```yaml
# elements/freedesktop-sdk.bst
track: freedesktop-sdk-26.08*
ref:   freedesktop-sdk-26.08.0-0-gdb97cce32cecadc7a3e98f06d557ebfa6ba9ad46
```

against `"extractVersion": "^freedesktop-sdk-(?<version>25\\.08\\.[0-9]+)$"`.

The description's framing of the failure mode is the important part and it's correct: an `extractVersion` matching nothing doesn't error, it reports no updates. "Silently stopped receiving bump PRs" and "already up to date" are indistinguishable from outside — so the thing that tells you the junction is current had itself gone blind.

**The guard actually catches it.** I reintroduced the `25\.08` regex and re-ran:

```
FAILED tests/unit/test_check_renovate_series.py::test_the_real_repo_is_consistent
1 failed, 5 passed
```

Clean tree: `6 passed`. It reads the real `renovate.json` against the real element, not a fixture of itself — so it will catch the next series bump too, which is the whole point.

**And it's wired to fire.** The `unit-tests.yml` path filter gains `renovate.json` and `elements/*.bst`. Without that the guard would live in a job a `renovate.json`-only change never triggers — a gate that exists but never runs, which is how the original drift survived in the first place.

- Full suite on this head: **207 passed, 1 xfailed** (baseline `main`: 201 passed, 1 xfailed)
- `mergeable_state: clean`, `build` and `unit` green

### One thing to flag before merging

The checker is reachable from pre-commit and from the test suite, but not from `just validate`, and there's no pre-commit CI workflow — so the unit test is the *only* enforcement in CI. Fine today; worth knowing it's a single point of failure if the path filter is ever pruned.

@castrojo @mrbobbytables — this is clean and needs no approval to merge under current protection, but it's Claude-authored end to end and nobody has reviewed it. Worth a human glance before it lands rather than merging on the strength of its own test suite.

The `just show-me-the-future` / `just install-vm` observation at the bottom of the description is worth its own issue, by the way — `main` having no automated boot gate is a bigger gap than this PR.

---

castrojo (approved):
Approving, after pushing a one-line fix. Explaining it since I am not the author, and flagging one caveat that outlives this PR.

**The bug is real and I verified it against `main` rather than the description.** `renovate.json` on `main` carries `"extractVersion": "^freedesktop-sdk-(?<version>25\\.08\\.[0-9]+)$"` while `elements/freedesktop-sdk.bst` declares `track: freedesktop-sdk-26.08*` with `ref: freedesktop-sdk-26.08.0-0-gdb97cce`. The filter targets a series the junction no longer tracks, so it matches nothing.

The guard is non-vacuous and correctly scoped: `extractVersion` applies to tag names returned by the `git-refs` datasource, not to the `ref:` value, and the `$` anchor deliberately excludes the git-describe `-0-g<sha>` suffix that `project.conf`'s `ref-format: git-describe` forces onto the pin. I read the checker line by line — `digits()` strips `\.` and `[0-9]+` before extracting groups, so `freedesktop-sdk-26.08*` -> `['26','08']` matches the corrected regex's `['26','08']`, and `gnome-50` -> `['50']` matches `^(?<version>50\.[0-9]+)$`. Comparing digits only means it would not catch a same-digits wrong-project regex, but the depName-to-junction mapping is explicit, so that is an acceptable bound and consistent with the sibling checkers.

**What I changed** (commit `312e11d`, `2fd5283..312e11d`, one line in `Justfile`): added `python3 .github/scripts/check-renovate-series.py` to the `validate` recipe, after its two siblings.

The new invariant was enforced only by pre-commit and pytest, while `check-release-version.py` and `check-k0s-version.py` both run under `just validate`. Since `AGENTS.md` and `docs/skills/ci-tooling.md` point contributors at `just validate` as the merge-contract check, a guard outside it is a guard people will not run. You flagged this yourself ("the unit test is the only enforcement in CI"), so I took it as an in-scope gap rather than a design change.

**Proof it actually guards**, which matters more than that it runs: temporarily reverting `extractVersion` to the 25.08 series makes `just validate` exit 1 with

    ERROR: elements/freedesktop-sdk.bst tracks `freedesktop-sdk-26.08*` (series 26.08) but renovate.json filters ... extractVersion '^freedesktop-sdk-(?<version>25\.08\.[0-9]+)$' (series 25.08). Renovate matches no tag, so this junction gets no update PRs.

Restored afterwards; `git status --porcelain` showed only `Justfile`. Clean run reports `OK: 2 Renovate version filter(s) match their junction track series` alongside the other two checkers.

**Verified at the new head:** `build` pass (44m33s), `build-kernel` pass, `unit` pass. `installer-test` is the pre-existing `main` failure and unrelated.

**One caveat worth recording, because this PR does not close it and should not be expected to.** Fixing the series is necessary but may not be sufficient for Renovate to actually open FSDK PRs. The custom manager's `matchStrings` capture `currentValue` from the `ref:` line — i.e. the git-describe string `freedesktop-sdk-26.08.0-0-gdb97cce...` — which cannot satisfy the anchored `extractVersion` either. Consistent with that, a repo-wide search for `freedesktop-sdk in:title` returns exactly one PR: this one. There is no evidence Renovate has ever opened an FSDK bump here. That deserves its own issue; it is out of scope for a series fix.

Related: I have opened #180, which moves junction tracking out of the `pull_request` path onto a scheduled workflow with its own PR. That is what currently delivers FSDK bumps in practice — via `track-refs` injecting them into unrelated Renovate PRs, which is the behaviour #180 removes.

Also correcting one line in your description, no action needed: "the `track-refs` job in `build.yml` never gets a ref to resolve" is not right — `track-refs` re-resolves the `track:` glob independently of Renovate and has been bumping FSDK all along. That is precisely how #49 and #176 acquired their undisclosed junction hunks.

Approving. You are the author so you cannot self-approve; this needs one more independent approval.

castrojo:
Reviewed and approved for merge — **blocked only on an OAuth token scope**, not on content.

`gh pr merge` refuses this PR with:

> refusing to allow an OAuth App to create or update workflow `.github/workflows/...` without `workflow` scope

The maintainer token currently carries `admin:public_key, gist, read:org, repo`. One command unblocks it:

```
gh auth refresh -s workflow
```

Merge order matters here, because these four collide in `build.yml` and `unit-tests.yml`. Verified locally by merging all four in sequence into a scratch `main`:

1. **#180** — biggest restructure; deletes the `track-refs` job and adds `track-junctions.yml`.
2. **#190** — conflicts with #180 on the deleted `track-refs` job. Resolution: keep the deletion, and pin `tool: just@1.58.0` in the **new** `track-junctions.yml` as well, which #190 could not know about.
3. **#189** — clean.
4. **#84** — clean.

After that sequence the merged tree passes: `actionlint` clean, `pytest tests/unit` 284 passed / 1 xfailed, `bats tests/unit` 91 passed, `docs-checks.py`, `check-release-version.py`, `check-renovate-series.py`, and `just validate`. All four `taiki-e/install-action` call sites end up pinned.

A direct push of that merge is refused by the `main` ruleset (`pull_request` required), which is correct — so these have to go through the API with the right scope.

PR #91, opened 2026-09-11
[sec-check] fix: generate kiosk TLS cert and key at first boot instead of sysext build time

### LINKED ISSUE #87: [architect] kiosk TLS private key baked into k0s sysext image at build time
## Architecture Finding

**Type**: anti-pattern / security
**Affected area**: `elements/oci/k0s-sysext.bst`

The k0s sysext build generates a self-signed TLS certificate for the KubeStellar kiosk proxy at **build time**:

```bash
openssl req -x509 -nodes -days 3650 -newkey rsa:2048 \
  -keyout sysext/usr/share/k0s/kiosk/key.pem \
  -out sysext/usr/share/k0s/kiosk/cert.pem \
  -subj "/CN=KubeStellar Console" \
  -addext "subjectAltName=DNS:localhost,DNS:*.local,IP:127.0.0.1,IP:10.0.2.15"
```

The private key (`key.pem`) is baked into the sysext image, which is published as a public GitHub Release asset. Every deployment from the same release shares the same TLS private key and certificate.

## Impact

- **Any party** who downloads the release artifact has the kiosk TLS private key
- The cert has a **10-year expiry** — rotation requires a full sysext rebuild and redeployment
- TLS provides no actual authentication since the key is not secret
- The cert includes `IP:10.0.2.15` (QEMU user-networking default), suggesting it was designed for local dev but shipped to production

## Recommendation

Move TLS certificate generation from build time to first boot:

1. Remove the `openssl req` command from `k0s-sysext.bst`
2. Add a `k0s-kiosk-tls.service` oneshot unit that generates the cert+key on first boot (before `k0scontroller.service`), using the host's actual IP addresses in the SAN
3. Store the generated key with `0600` permissions on the host's `/var` partition (persistent, per-host)

This aligns with the existing pattern in server#80 where SSH host keys moved from build-time to first-boot generation.

---
*Filed by architect agent (ACMM L5 — hold-gated mode)*
---
🐝 **Hive Agent**: `architect` | **Instance:** `hosted-projectbluefin-knuckle-gjvq` | **SHA:** `unknown`

— hive: agent=architect backend=copilot model=kimi-k3

Danathar:
## Security Fix

Addresses architecture finding in #87: the k0s sysext build previously generated a self-signed TLS certificate and private key (`key.pem`) at build time and baked them directly into the published sysext image.

### Changes

1. **Remove Build-Time TLS Generation**: Removed `openssl req` and `freedesktop-sdk.bst:components/openssl.bst` build-depends from `elements/oci/k0s-sysext.bst` so no private key or static certificate is baked into the public sysext release asset.
2. **First-Boot Oneshot Service**: Added `files/k0s/sysext/k0s-kiosk-tls.service`, a oneshot systemd unit packaged into the sysext that generates `key.pem` (mode `0600`) and `cert.pem` (mode `0644`) under persistent `/var/lib/k0s/kiosk/` on first boot before `k0scontroller.service` starts.
3. **Dynamic SAN Resolution**: Populates Subject Alternative Names with `DNS:localhost`, `DNS:*.local`, `IP:127.0.0.1`, hostname, and all global IP addresses discovered on the host (`ip -o addr show scope global`).
4. **Idempotence and Rotation**: Skips generation if both `key.pem` and `cert.pem` exist. Operators can rotate the certificate at any time by removing them from `/var/lib/k0s/kiosk/` and restarting the unit.
5. **Controller Ordering**: Added `After=network-online.target k0s-kiosk-tls.service` and `Wants=network-online.target k0s-kiosk-tls.service` to `files/k0s/sysext/k0scontroller.service`.
6. **OpenSSL Runtime Dependency**: Added `freedesktop-sdk.bst:components/openssl.bst` to `elements/bluefin-server/os-stack.bst` to provide `/usr/bin/openssl` at runtime.
7. **Documentation & Unit Tests**: Updated operator runbook in `docs/skills/k0s-sysext-ops.md` and repository layout in `docs/skills/k0s-sysext.md`. Added unit test contracts in `tests/unit/test_kubestellar_kiosk.py`.

Closes #87

— hive: backend=agy


hanthor (changes requested):
## The vulnerability fix is right. The "persistent" half does not hold.

Removing the build-time `openssl req` is correct and worth doing — a private key baked into a published sysext is the same key on every install, and that is a real finding. No argument with the direction.

But the certificate this PR generates is **regenerated on every boot**, not persisted, because nothing removed the tmpfiles rule that owns the destination directory.

### The collision

`files/k0s/sysext/k0s-manifests.conf` is unchanged by this PR and still contains:

```
C+ /var/lib/k0s/kiosk - - - - /usr/share/k0s/kiosk
```

The new unit writes into that exact path:

```
ExecStart=/bin/bash -eu -c 'test -s /var/lib/k0s/kiosk/key.pem && test -s /var/lib/k0s/kiosk/cert.pem && exit 0; mkdir -p /var/lib/k0s/kiosk; ... -keyout /var/lib/k0s/kiosk/key.pem -out /var/lib/k0s/kiosk/cert.pem ...'
```

In `tmpfiles.d`, the `+` suffix on `C` means *remove the destination first, then copy*. So on every boot:

1. `systemd-tmpfiles-setup.service` runs (ordered early, before `sysinit.target`) and **deletes `/var/lib/k0s/kiosk/` wholesale**, re-copying it from `/usr/share/k0s/kiosk` — which, after this PR, no longer contains `cert.pem` or `key.pem`.
2. `k0s-kiosk-tls.service` runs much later (`After=network-online.target`, well past `sysinit.target`), finds no key, and its `test -s … && exit 0` fast path falls through.
3. A brand-new self-signed cert is minted.

The unit is named "Generate **persistent** TLS certificate and key", the docs added here say the key "is stored **persistently** in `/var/lib/k0s/kiosk/key.pem`", and the idempotence guard in item 4 of your summary is written to skip regeneration — none of those are true as long as that `C+` line points at the same directory. Operators get a new certificate fingerprint on every reboot, so any pinned or hand-trusted cert breaks and the browser warning returns each time.

Note this is *new* with this PR: before it, `cert.pem`/`key.pem` shipped inside `/usr/share/k0s/kiosk`, so the `C+` re-copy reproduced the same (shared, baked-in) cert each boot. That was the vulnerability — but it was at least stable. This PR fixes the sharing and trades it for churn.

### Suggested fix

Cleanest is to stop overlapping the two owners. Generate into a directory tmpfiles does not manage, e.g. `/var/lib/kubestellar-kiosk/`, and point the proxy's hostPath at it:

```yaml
# files/k0s/manifests/kubestellar/41-kubestellar-kiosk-proxy.yaml
          path: /var/lib/kubestellar-kiosk
```

(`nginx.conf` needs no change — it reads `/etc/kubestellar-kiosk/cert.pem`, which is the mount point, not the host path.) Dropping the `+` from the kiosk line would also stop the deletion, but then sysext upgrades stop propagating new kiosk assets, which is presumably why `+` is there.

Whichever way you go, please add ordering — `After=systemd-tmpfiles-setup.service` — so the two can never race, and a test asserting the generated path is not under a `C+` destination.

### Smaller points

- **`ip` / `hostname` may be absent.** The SAN loop is guarded by `command -v`, so if `ip` is not in the OS stack the cert silently gets `DNS:localhost,DNS:*.local,IP:127.0.0.1` and no host IP — the console is then unreachable by address without a cert error, with nothing in the log saying why. `openssl` was correctly added to `os-stack.bst`; consider asserting `ip` is there too, or logging when the lookup yields nothing.
- **`After=network-online.target`** delays `k0scontroller.service` (via `Before=`) purely to enumerate addresses. That is a real boot-time cost on a host with a slow link, for a cert that this PR intends to generate once.
- **First-boot SAN is frozen for 3650 days.** On DHCP the address can move and the cert does not follow. Worth documenting in the runbook next to the rotation steps you added.

### CI status

**No checks have run on this PR at all:**

```
mcp__github__pull_request_read(method="get_check_runs", pullNumber=91)
  -> {"total_count":0,"check_runs":[]}
```

`mergeable_state` is `unstable`. It is a fork PR (`Danathar/server`) and appears to need maintainer approval to run workflows. The build job is the only thing that would exercise `k0s-sysext.bst` after the `openssl` build-depend was removed, so that removal is currently unverified by CI. I ran the local gates on `main + this branch` and they pass — `204 passed, 1 xfailed`, 41/41 bats, all three checkers exit 0 — against a `main` baseline of `201 passed, 1 xfailed`, so the three new tests are additive and nothing regressed. But the local suite cannot build the sysext.

### Merge-order note

This and #90 both edit `docs/skills/k0s-sysext-ops.md` and `tests/unit/test_kubestellar_kiosk.py`. Each merges cleanly into `main` on its own, but they conflict with **each other**:

```
$ git merge pr90   # onto main+pr91
CONFLICT (content): Merge conflict in docs/skills/k0s-sysext-ops.md
```

Whichever lands second needs a rebase.

---

Danathar:
@hanthor — pushed in `6c2a436`, but with a correction on the mechanism first.

**I couldn't reproduce the "deletes wholesale every boot" claim.** I ran `systemd-tmpfiles --create --remove --boot --exclude-prefix=/dev` (the exact invocation `systemd-tmpfiles-setup.service` uses) twice against a sandboxed root using the *original* single-line `C+ /var/lib/k0s/kiosk - - - - /usr/share/k0s/kiosk` config, planting fake `cert.pem`/`key.pem` between the two runs to simulate a reboot after cert generation. On this host's systemd 259 (Fedora 44, matching the target OS), they survived both runs.

Checked the actual semantics (upstream docs 403'd, used the Arch Wiki mirror): plain `C` skips the *entire* copy if the destination directory already exists and is non-empty; `C+` only relaxes that skip — it still doesn't delete or overwrite files already present at the destination. So "removes the destination first, then copies" isn't accurate for this systemd version — `C+` fills in what's missing, it doesn't wipe first.

**Made the fix anyway.** Regardless of whether the exact mechanism reproduces, relying on an undocumented, version-dependent "`C+` happens to leave extra files alone" behavior as the only thing between a private key and deletion is fragile for something this sensitive — and you asked for the ordering guarantee either way. So:
- Split the one whole-directory `C+` into three per-file lines (`nginx.conf`, `kiosk-gate.js`, `kiosk-gate.css`) — `cert.pem`/`key.pem` are now structurally never named by any tmpfiles directive, not just incidentally spared by current behavior.
- Added `After=systemd-tmpfiles-setup.service` to the ordering.
- Added logging when `hostname`/`ip` are unavailable (your smaller point).
- Added a docs note on SAN staleness under DHCP.
- Added a regression test asserting no tmpfiles line targets the bare directory and that `cert.pem`/`key.pem` never appear in the config at all.

Full suite: 205 passed, 1 xfailed (unchanged).

Left the `network-online.target` boot-time-cost point undone — that's a real tradeoff question (probe delay vs. waiting), not a bug, and felt like a separate decision from this PR's scope.

castrojo:

The branch could not be refreshed onto current `main` automatically because `elements/oci/k0s-sysext.bst` conflicts. Please rebase and resolve that file. Current `main` includes #149, so the refreshed branch will use the shared Project Bluefin QEMU installer gate.

castrojo:

Update: the second refresh onto `main` completed successfully after the shared KVM fix landed. No author action is needed for the branch refresh.

castrojo (commented):
Reviewed. The blocking objection was answered in code two commits ago, so the CHANGES_REQUESTED here is stale. Holding my approval only until `build` reports, for a specific reason given below.

**First: CI had never run on this PR at all.** That is why it looked stalled. This is a fork PR from a contributor with `read` access, and `build.yml` triggers on plain `pull_request` while checking out the PR head, so every run sat at `conclusion: action_required` with zero jobs dispatched — indistinguishable from "no checks yet" in the UI. I have approved the runs. `unit` and `docs` now pass and `build` is in progress.

**@hanthor's blocking objection is addressed.** His concern was that the `C+ /var/lib/k0s/kiosk` whole-directory tmpfiles rule would wipe the generated material each boot, so the unit's `test -s ... && exit 0` fast path would fall through and mint a fresh cert every boot — making the "persistent" claim in the unit description, the docs and the idempotence bullet false. The current head addresses all three things he asked for: the whole-directory rule is split into three per-file `C+` lines for `nginx.conf`, `kiosk-gate.js` and `kiosk-gate.css`, so the generated TLS material is never named by tmpfiles; the unit gained `After=systemd-tmpfiles-setup.service`; and there is a `test_tmpfiles_never_wipes_the_generated_tls_material` test. I verified all three are present in the current diff. @hanthor, this is 2 commits newer than the head you reviewed — worth a fresh look.

The security direction is right: generating the cert and key at first boot rather than at sysext build time means machines stop sharing one baked-in key, which is the whole point.

**Why I am waiting on `build`.** This deletes the `openssl` build-depend from `elements/oci/k0s-sysext.bst` and moves generation into a first-boot unit. Removing a build-depend from a `.bst` element is exactly the class of change only the `build` job can validate — `docs/skills/avoid-over-engineering.md` documents the failure mode where a dropped dependency leaves the sandbox without a tool and a `script`/`manual` element fails in a confusing way. Since that job has literally never executed on this branch, approving now would be approving on no evidence for the one thing most likely to break. I will approve once it is green.

**One live review item, not a stale one.** `files/k0s/sysext/k0s-kiosk-tls.service` combines `After=network-online.target` with `Before=k0scontroller.service`, which delays Kubernetes startup behind `network-online` on *every* boot — including the overwhelmingly common case where the cert already exists and the unit does nothing but run two `test -s`. @hanthor raised it; the reply was that it is "a real tradeoff question... felt like a separate decision". I do not think it is separate, because the cost is paid on every boot forever while the benefit applies only to the first. That said, it is a judgement call about ordering semantics rather than a correctness bug, and it is the author's to make — I am flagging it, not blocking on it.

**Two smaller things.** `[Install] WantedBy=k0scontroller.service multi-user.target` in that unit is dead weight: the sysext ships no enablement symlink and `k0scontroller.service` already carries `Wants=k0s-kiosk-tls.service`. Harmless, but it makes a reader believe an enablement path exists. And the SAN is frozen for 3650 days at first boot, so on DHCP the address moves and the cert does not follow — correctly disclosed in the runbook rather than hidden, which I am fine with.

**On the new tests.** The added assertions are source-text greps (`assert "keyout" not in sysext`, exact tmpfiles line strings). That is the same pattern `main` recently deleted `test_vm_dashboard_contract.py` for. It is consistent with the rest of that file so I am not making it a blocker here, but it is worth not extending further — a test that greps the element text cannot catch the build-sandbox breakage that `build` is currently checking for.

No security defect found in the shell: `$$(hostname)` and `awk "{print $$4}"` are systemd's literal-dollar escape, so bash receives `$(hostname)` and awk receives `{print $4}` as intended.

**Conflicts to sequence, not to fix now.** This and #90 both touch `docs/skills/k0s-sysext-ops.md` and `tests/unit/test_kubestellar_kiosk.py`; each merges into `main` alone but not into the other, so whichever lands second rebases. This and #80 both edit `elements/bluefin-server/os-stack.bst` about three lines apart — probably auto-merges, worth eyeballing.

Not approving yet, purely pending `build`. Everything else here is ready.

Danathar:
Pushed `27b76f1` for the SAN finding above. It was real: because the ExecStart program is a single-quoted `bash -eu -c` string, the awk program had to be double-quoted, so after systemd rewrote `$$` to `$`, bash expanded `$4` itself, `-u` killed the command substitution, and the address loop ran zero times.

What changed:

- The field extraction is now `tr -s " " | cut -d " " -f4 | cut -d/ -f1`, which needs no `$` at all.
- Two behavioral tests run the exact ExecStart program (with the `$$` rewrite applied) against a fake state directory with stubbed `ip`, `hostname`, and `openssl`. One asserts the IPv4 and IPv6 addresses and hostname land in `subjectAltName` and the files get 0600/0644; the other asserts an existing key and cert are left alone. The first test fails on the previous unit and passes now. Full unit suite: 213 passed locally.

@castrojo this will have dismissed nothing of yours since you had not approved yet; the `installer-test` failure is separate and I'm looking at it next.

Danathar:
On `installer-test`: the failure on `35d368e` is not this PR. The QEMU boot logs `Timed out waiting for device /dev/disk/by-partlabel/var`, every unit that needs `/var` fails as a dependency, and the console readiness wait then times out after 900s. Every `Build DDI artifacts` run in the repo since 2026-09-15 fails the same way, including `main`. mrbobbytables has #173 open for the `/var` mount deadlock; once that lands this PR's build should be re-run rather than read as a regression here.

castrojo:

\`elements/oci/k0s-sysext.bst\` and \`files/k0s/\` are deleted. #210 fixes #87 by removing the build-time key entirely rather than regenerating it at first boot: kiosk TLS is issued by **cert-manager** at runtime (\`files/cluster/manifests/pki/issuers.yaml\` + \`files/cluster/manifests/kiosk/certificate.yaml\`).

That is strictly better than a first-boot generator script — the key material is managed, rotatable, and never passes through a release asset. Your diagnosis of the leak was correct and it is what drove the cert-manager dependency.

Issue #96, opened 2026-09-12
Confirm the curl component path in the freedesktop-sdk junction


castrojo:
## What

`elements/bluefin-server/os-countme.bst` adds two runtime dependencies:

```yaml
- freedesktop-sdk.bst:components/curl.bst
- freedesktop-sdk.bst:components/jq.bst
```

`components/jq.bst` is confirmed — projectbluefin/dakota already depends on it by that exact path. `components/curl.bst` is **not confirmed**: no repo in the factory references curl through this junction, so the path was inferred rather than verified.

## What fixed looks like

Either the path resolves and this closes, or it is corrected to whatever freedesktop-sdk actually calls it:

```bash
bst show elements/bluefin-server/os-countme.bst
```

## Why it matters here specifically

This image is minimal and ships neither curl nor jq today. The reporter reads image metadata with jq and sends the ping with curl, so a wrong path is not a cosmetic failure — the element fails to resolve and the image does not build.

## Impact

Caught by the first CI run on #94. Filing so it is not lost if that run is not read closely.

mrbobbytables:
Confirmed: `components/curl.bst` is the exact path in freedesktop-sdk (`elements/components/curl.bst` under element-path `elements`). The dependency `freedesktop-sdk.bst:components/curl.bst` in PR #94 is correct and resolves successfully in the `Validate Element Graph` CI step.
---
🐝 **Hive Agent**: `contributor` | **SHA:** `9428435`

castrojo:
Confirmed: curl component path in freedesktop-sdk junction is elements/components/curl.bst, matching components/curl.bst. Locked in via #122.

castrojo:
Confirmed and now gated. The pinned junction (`freedesktop-sdk-26.08.0`, `elements/freedesktop-sdk.bst`) ships both `components/curl.bst` and `components/jq.bst`, and `elements/bluefin-server/os-countme.bst` depends on exactly those paths.

#122 locks it in: `test_os_countme_depends_on_curl_and_jq` fails if either dependency is renamed or dropped, so the minimal image cannot silently lose the tools the countme reporter needs.

PR #122, opened 2026-09-13
test(build-depends): lock in os-countme curl+jq FSDK paths

kylerankin:
## Issue: projectbluefin/server#96

Confirm the curl component path in the freedesktop-sdk junction.

## What was verified

`elements/bluefin-server/os-countme.bst` declares:

```yaml
- freedesktop-sdk.bst:components/curl.bst
- freedesktop-sdk.bst:components/jq.bst
```

`jq.bst` was already confirmed. For `curl.bst` I checked the pinned junction
ref directly instead of trusting the inference. The junction
(`elements/freedesktop-sdk.bst`) pins `freedesktop-sdk-26.08.0-0-gdb97cce32cecadc7a3e98f06d557ebfa6ba9ad46`, whose element-path is `elements/`:

```bash
git ls-tree -r --name-only db97cce32cecadc7a3e98f06d557ebfa6ba9ad46 \
  | grep -E 'components/(curl|jq)\.bst$'
# elements/components/curl.bst
# elements/components/jq.bst
```

Both resolve at the pinned ref, on the exact same `freedesktop-sdk.bst:components/<X>.bst`
path that `ca-certificates`, `systemd`, `podman`, and `jq` already use in this
repo. The path is correct and resolves.

## Change

The minimal image ships neither curl nor jq, so a wrong junction path fails to
resolve and the image does not build. This adds a regression test
(`test_os_countme_depends_on_curl_and_jq`) locking the two dependency paths so
the confirmed resolution cannot silently drift.


hanthor (approved):
Verified: `elements/bluefin-server/os-countme.bst` does declare `freedesktop-sdk.bst:components/curl.bst` and `freedesktop-sdk.bst:components/jq.bst`, so the test asserts something true rather than aspirational. `pytest tests/unit` 212 passed / 1 xfailed, `bats` 62 ok.

**One coordination note, not a change request.** #132 already contains this exact test, copied verbatim into the slot where it replaces `test_os_stack_includes_bash`. Merging both into `main` conflicts on `tests/unit/test_build_depends.py`, and a careless resolution leaves two `def test_os_countme_depends_on_curl_and_jq` in one module — **Python keeps the second silently**, so the duplicate never surfaces as an error.

Cheapest fix: land this one first (it's one function and no risk), then have #132 rebase and drop its copy. I've noted the same on #132.

**Separately, worth knowing while #132 is open.** Flatcar 4593.2.5's `/usr` ships `/usr/bin/curl` and `/usr/bin/jq` — I checked the tarball listing. So if #132's cutover lands, `os-countme.bst` would be pulling FSDK-linked curl/jq onto a Flatcar glibc, which is exactly the two-ABI situation #132 says it's removing. This test would still pass while that's true, which is worth being explicit about in its docstring: it pins the *declaration*, not the ABI.

---

kylerankin (commented):
Thanks for the review and the merge, Jorge.

On the ABI note: good catch. This test pins the *declaration* (the FSDK junction path resolves on the pinned FSDK ref), not the ABI — on Flatcar 4593.2.5's `/usr` ships its own curl/jq, so the assertion would still pass even though the ABI situation #132 is removing is exactly what makes the FSDK-linked curl/jq undesirable there. The docstring already states it verifies the pinned ref ships both components; I'll make the "declaration, not ABI" distinction explicit in #132's copy of this test, since that's the version that lands next to the Flatcar cutover where the question is live. For this PR (FSDK-only), the current wording is accurate as-is, so I'm leaving the approved test untouched rather than resetting the approval for a doc-only tweak.

Sequencing noted: land #122 first, then #132 rebases and drops its duplicate copy of `test_os_countme_depends_on_curl_and_jq`.

castrojo (approved):
Reviewed the diff and verified the claim independently rather than taking the PR body at face value.

Verification performed:
- Checked out the head (e3f3d9c) and ran `python3 -m pytest tests/unit/test_build_depends.py -q`: 7 passed.
- Confirmed against `main` that `elements/bluefin-server/os-countme.bst` really does declare both `freedesktop-sdk.bst:components/curl.bst` and `freedesktop-sdk.bst:components/jq.bst`, so the new assertion locks an invariant that currently holds rather than encoding an aspiration.

The test earns its place: `os-countme.bst` is the only consumer of curl and jq, the minimal image ships neither, and a wrong junction path fails at build time with a resolution error rather than anything obvious. Pinning the two dependency paths is a real regression guard against silent drift, and it asserts an observable build contract rather than implementation detail. Scope is one test function added to an existing file, matching the surrounding style.

CI note: the only failing check is `installer-test / Install and boot in QEMU`. That job also fails on `main` at tip 5ebfeae (run 34920225526), where the installed system drops to emergency mode with "Dependency failed for /var" and "Dependency failed for Local File Systems" before timing out on KubeStellar Console readiness. That failure is upstream of this PR and unrelated to it; `build` and `unit` are green here.

Approving. Merge should still wait until the `/var` regression on main is fixed so this lands on a green base.

kylerankin:
Thanks for the coordination note — good catch. Dropped the verbatim `test_os_countme_depends_on_curl_and_jq` copy from #132 (`dd31769`); #132 keeps only its own new `test_installer_linker_paths_split_host_and_target`. So #122 can land first with the curl/jq test and #132 rebases onto it with no duplicate-on-merge risk.

And thanks for the heads-up on the Flatcar 4593.2.5 `/usr` curl/jq two-ABI point — noted for when the cutover actually lands; worth a dedicated look rather than something this test would catch.
---

PR #123, opened 2026-09-13
docs(spec): Flatcar version parity on the FSDK base

castrojo:
## Summary

Design for **version parity** with Flatcar Container Linux: for each component Flatcar ships, build that same upstream version from source in BuildStream, on the freedesktop-sdk 26.08 base.

The question the project answers: **what would Flatcar look like if it were built with BuildStream?**

## Version parity, not ABI parity

- **Version parity** — systemd 257 means systemd 257; kernel 6.12.102 means 6.12.102.
- **ABI parity is explicitly not the goal.** Flatcar's kernel banner reads `x86_64-cros-linux-gnu-gcc (Gentoo Hardened 14.3.1_p20250801 p4)`. Our binaries will differ. Accepted.

Done is not "our bytes match theirs". Done is "we build the versions Flatcar builds, the result boots and behaves, and we can say precisely where a BuildStream-built Flatcar diverges from the real one."

**Hard rule 1 is unchanged.** The base stays FSDK 26.08. Flatcar binaries are imported as a *reference* to boot against and measure, never as the shipped payload.

## Versioning

```
26.08.XX.$FLATCARVERSION      e.g. 26.08.13.4593.2.5
```

Both halves enforced: FSDK point release against the junction ref, Flatcar release against the pin in `include/flatcar.yml`.

## Method: import first, then substitute

Parity needs an oracle. Import Flatcar's binaries, boot them on our installer and layout to establish a known-good control, then replace components one at a time with FSDK-built ones at Flatcar's versions, re-running the boot proof after each swap.

## Boot proof (already achieved)

Built a conforming disk from upstream artifacts alone, unprivileged — `sfdisk`, `mksquashfs`, `mke2fs -d`, no root, no loop mounts — and booted the pinned kernel with no external initrd:

```
[    1.122448] Run /init as init process
Mounting /usr from /dev/vda2
[    1.776610] systemd[1]: Successfully made /usr/ read-only.
[    4.133550] systemd[1]: Switching root.
Welcome to Flatcar Container Linux by Kinvolk 4593.2.5 (Oklo)!
[  OK  ] Reached target multi-user.target - Multi-User System.
localhost login:
```

## Findings that drove the design

- `CONFIG_INITRAMFS_SOURCE="bootengine.cpio"` — the kernel already carries a two-stage initramfs. The installer has been generating a second one to lay over a working one.
- `veritysetup --hash-offset=1065345024` is hardcoded in stage 1: verity is **one** partition with the hash appended, not systemd-repart's two-partition model. `/usr` has a hard 1,065,345,024-byte budget.
- Stage 2 is `bootengine.img`, loop-mounted from the mounted `/usr` — so it versions with the payload, and must survive the import.
- `OEM` is **required**, matched by *filesystem* label; without it the boot stalls 90s then drops to emergency.
- Ignition needs no masking — it degrades to `Subsequent (Not Ignition) boot complete`.
- A/B is `USR-A`/`USR-B` (`5dfbf5f4-2848-4bac-aa5e-0d9a20b745a6`), not `root-a`/`root-b`.

## Scope

Design only, no element changes. Ticket chain: #125 versioning · #137 pins · #142 parity matrix · #126 reference import · #134 layout · #135 verity · #128 dracut removal · #129 comparison harness · #130 boot proof in CI · #138 creds · #139 reboot coordination.

Assisted-by: GPT-5.6 Sol via GitHub Copilot


hanthor (changes requested):
This document contradicts the four implementation PRs open against it, and in places contradicts itself. As written it can't serve as the design of record for #132/#133/#136/#140/#141.

## 1. Hard rule 1: stated as unchanged, three times

> **The build base stays freedesktop-sdk 26.08.** Hard rule 1 is unchanged: compose from FSDK `components/*`. Version parity is pursued *within* the FSDK graph, by building the versions Flatcar ships rather than by importing Flatcar's tree wholesale.

> Both images are composed from FSDK 26.08. […] **Installed OS payload** tracks Flatcar's versions […] still built from FSDK `components/*`.

> Hard rule 1 is untouched throughout: the base stays FSDK 26.08.

#133 changes that rule to "Installer image composes from FSDK 26.08 `components/*`; OS payload may import Flatcar binaries pinned by digest," and its ADR explicitly rejects the hybrid this document describes: *"Keeping FSDK userspace in the OS payload while importing Flatcar kernel and ZFS artifacts preserves the ABI split that the migration is meant to remove."*

## 2. Internally inconsistent

The paragraph above says "rather than by importing Flatcar's tree wholesale." Two sections later, `### New element: flatcar/flatcar-usr.bst` specifies exactly that import, and `### Overlay: what Flatcar displaces` tabulates the removal of the entire FSDK base runtime. Phase 3 then says the imports are *"the oracle, not the payload"* — but #132's `os-stack.bst` makes `flatcar-usr.bst` the payload.

These are two different designs in one file, and the reader can't tell which one governs.

## 3. ABI parity

This says *"ABI parity is explicitly not the goal […] the binaries will differ, and that is expected and accepted."* #132's element header says it *"Replaces the FSDK 26.08 base userspace […] with a **single-ABI** Flatcar `/usr`."* Opposite objectives.

## 4. §Versioning explicitly supersedes #136

> `.github/scripts/check-release-version.py` **inverts. It stops enforcing `release-version` against the FSDK junction ref** and starts enforcing it against `flatcar-version`.
>
> This supersedes both earlier proposals: **the two-field split** and the `26.08.XX.$FLATCARVERSION` composite.

#136 *is* the two-field split (`installer-version` + `flatcar-version`) and it keeps enforcing `installer-version` against the junction ref. One of the two has to be withdrawn.

For what it's worth I think #136 has the better of the argument — this document's own "the installer and the payload diverge in version" section is an argument *for* two axes, since `systemd-sysinstall` only exists in the installer's FSDK systemd and the installer genuinely cannot follow the payload's version. But that reasoning supports two axes, not collapsing onto one.

## Two things here the implementations should adopt, and one that's factually wrong

**Adopt:** *"Removals are explicit `rm` lines with a comment naming the replacement, not a wildcard sweep, so a future Flatcar bump that renames a unit fails loudly."* #140 follows this exactly. #132's single `rm -rf` block does not — and I confirmed against the real tarball that three of its paths don't exist and fail silently, while six real ones go unremoved.

**Adopt:** *"Contract tests assert `/usr/lib/flatcar/bootengine.img` is present and the module layout is flat."* #140 has these; #132 doesn't.

**Wrong:** the design (and both implementations) assume Flatcar nests modules at `usr/lib/modules/<kver>/<kver>/`. I listed all 19,658 entries of `flatcar-container.tar.gz` (sha256 `c533ff57…`, the ref both PRs pin) — **the tree is flat.** There is no nested directory to collapse, so the flattening block is dead code in both. The likely origin of the belief is the `--strip-components=5` in the existing `flatcar-kernel.bst`.

## Also missing

The "what Flatcar displaces" table maps podman → `flatcar-podman.raw` sysext, but #141 additionally imports containerd and docker, which this document doesn't mention at all. And the document says the target `dracut` invocation "is deleted, and nothing replaces it" — none of #132/#140/#141 touches `elements/oci/bluefin-server-installer.bst`, where that invocation lives.

`docs-checks.py` passes and `pytest`/`bats` are unaffected — this is a documentation-consistency objection, not a test failure.

---

castrojo:
Thanks for the thorough review. The core tensions you identified are valid and reflect the ongoing discussion around the migration strategy:

1. **Hard Rule 1 & Scoping:** As established in PR #133 and its ADR (`docs/superpowers/specs/2026-09-13-flatcar-binary-imports-adr.md`), the intended architectural boundary is scoping FSDK 26.08 composition strictly to the installer image (`systemd-sysinstall`), while permitting the installed OS payload to import digest-pinned Flatcar binaries (`flatcar-usr.bst`, sysexts). We will update this spec to align directly with #133 rather than asserting that the OS payload rebuilds Flatcar components from source within FSDK.
2. **Version Parity vs. Single-ABI Import:** The spec attempted to reconcile two competing frames ("build Flatcar versions in FSDK" vs "import Flatcar /usr as a single-ABI baseline"). We agree the hybrid framing introduces contradictions. The document will be revised so that the binary import model (single ABI domain, matching #132/#140) is clearly stated as the accepted design, resolving the conflict.
3. **Module nesting & dead code:** Verified your finding on `flatcar-container.tar.gz` module layout being flat (no `<kver>/<kver>/` nesting). The dead flattening logic and inaccurate nesting claims are being excised.
4. **Versioning Axes & #136:** Acknowledged on the two-axis model (`installer-version` tracking FSDK and `flatcar-version` tracking the payload). We will align the versioning section with #136's two-axis approach.

castrojo:
Closing my own PR. This document cannot serve as the design of record in its current state, and rather than leave four dependent PRs anchored to a spec that says two incompatible things, I am withdrawing it and will rewrite it against the decisions that have actually been made since.

@hanthor raised five specific contradictions. I replied on 2026-09-14 saying I would address them. Twelve commits later, at head `33576b3`, every one of them is still present verbatim — I checked rather than trusting my own comment:

- "Hard rule 1 is unchanged" still at lines 33, 540 and 562, while #133 changes exactly that rule and its ADR explicitly rejects the hybrid this document describes.
- "ABI parity is explicitly not the goal" still at 16, 21 and 506, while #132's element header delivers a single-ABI Flatcar `/usr`.
- "rather than by importing Flatcar's tree wholesale" at 34-36 sitting in the same file as "### New element: `flatcar/flatcar-usr.bst`" at 175, which does exactly that.
- "This supersedes both earlier proposals: the two-field split and the 26.08.XX.$FLATCARVERSION composite" still at 483-484, while #136 *is* the two-field split.
- The module-nesting claim at 57-58 — that Flatcar nests modules at `usr/lib/modules/<kver>/<kver>/` — still present, and disproven: hanthor listed all 19,658 entries of `flatcar-container.tar.gz` at the pinned sha256 and the tree is flat. The same file then asserts "the module layout is flat" at line 577.

The PR body also contradicts the file it ships: it advertises `26.08.XX.$FLATCARVERSION` as the versioning scheme, which the spec itself says is superseded.

There is one more violation nobody raised. `docs/skills/index.md` carries the standing fact "Version scheme: FSDK-derived only; no separate application version axis." This spec introduces exactly such a second axis and changes only one file, leaving that canonical line standing — AGENTS.md rule 6.

**Why closing rather than revising in place.** Points (a) through (d) each require picking one of two contradictory designs — FSDK rebuild versus wholesale Flatcar import, one version axis versus two. Those are no longer open questions the way they were when I wrote this: #140 landed the wholesale-import element in reviewable shape, #133 rescopes rule 1 to permit it, and #136 implements the two-axis split. The honest move is to rewrite the spec to describe what we decided, not to patch a document that argues against the implementations now in flight.

**What this does not block.** Nothing depends on this merging. The four PRs that cite it — #148 (#142 parity matrix), #163 (#129 comparison harness), the now-closed #145 (#139 reboot coordination), and #133 — each stand or fall on their own contents, and I have reviewed them individually on that basis.

I will reopen a replacement spec once #133 and #136 settle the rule-1 and versioning questions, so it can record decisions rather than propose contradictory ones.

For the record, two things from hanthor's review that the implementations should keep regardless of what the rewritten spec says, because they have already proven their worth: explicit `rm` lines each naming the Bluefin component that replaces the removed Flatcar one, rather than a wildcard sweep — #140 does this and it is the right pattern; and contract tests asserting `/usr/lib/flatcar/bootengine.img` survives the import, since stage 1 loop-mounts it as stage 2.

Issue #126, opened 2026-09-13
feat(flatcar): import Flatcar /usr as a boot reference, not as the payload


castrojo:
Design: `docs/superpowers/specs/2026-09-13-flatcar-base-migration-design.md (PR #123)`, phase 3. Blocked by #142.

**Reframed by the version-parity decision.** This is no longer "replace the OS payload with Flatcar's tree". The base stays FSDK 26.08 and hard rule 1 is unchanged. Flatcar's binaries come in as the **oracle**: a known-good tree to boot on Bluefin Server's installer and partition layout, against which each FSDK-built component at Flatcar's version is later measured.

Building a component and asking "is this right?" is unanswerable without something known-good to compare to. That is what this element provides.

## Change

Add `elements/flatcar/flatcar-usr.bst`, `kind: manual`, mirroring `flatcar-zfs.bst`:

- One `kind: remote` source for `flatcar:stable/%{flatcar-board}/%{flatcar-version}/flatcar-container.tar.gz`, pinned by sha256 via the existing `flatcar:` alias.
- `variables: strip-binaries: ""` - prebuilt binaries must not be stripped.
- Extract `./usr` into `%{install-root}/usr`.
- Flatten Flatcar's nested `usr/lib/modules/<kver>/<kver>/` layout.
- **Preserve `usr/lib/flatcar/bootengine.img`.** Stage 1 loop-mounts it from the mounted `/usr`; removing it strands the boot between stages.

Remove Flatcar's update and provisioning stack with explicit `rm` lines, each commented with its Bluefin replacement: `update_engine`, `update_engine_client`, `update_engine_stub`, `locksmithd` + `locksmithd.service`, `flatcar-update`, `download_sysext`, `ensure-sysext.service`. No wildcard sweeps.

Note the Ignition binary and units can stay: the boot proof showed the state machine degrades to a no-op without a config (`ignition-subsequent.target - Subsequent (Not Ignition) boot complete`).

## Acceptance

- `just validate` passes.
- Contract test asserts the update stack is absent, `bootengine.img` is present, and the module tree is single-level.
- The element boots through the layout in #134, reproducing the recorded boot proof.
- Clearly marked in-tree as a reference artifact, not a shipped payload.

castrojo:
Constraint from the stage-1 analysis in #128: **`/usr/lib/flatcar/bootengine.img` must be preserved by the strip list.**

Stage 1 loop-mounts it from the `/usr` it just mounted:

```sh
[...]

PR #133, opened 2026-09-13
docs(arch): scope FSDK hard rule to installer

### LINKED ISSUE #124: arch: scope hard rule 1 to the installer so the OS payload can import Flatcar binaries
Design: `docs/superpowers/specs/2026-09-13-flatcar-base-migration-design.md (PR #123)`, phase 1.

AGENTS.md hard rule 1 reads "Compose from FSDK 26.08 `components/*`". The Flatcar base migration needs that scoped to the installer, with pinned, digest-verified Flatcar binary imports permitted for the OS payload - the model `flatcar/flatcar-kernel.bst` and `flatcar/flatcar-zfs.bst` already follow.

Blocks every element-level phase. Hard rules 2-6 are unaffected: the CPU baseline stays broad (Flatcar amd64-usr is x86-64 baseline), the installer stays `systemd-sysinstall`-native on FSDK, and boot entries keep using PARTUUID.

Acceptance:
- Hard rule 1 amended, naming the installer/OS-payload split explicitly.
- An ADR records the decision and the rejected alternatives (source rebuild, status quo hybrid).

castrojo:
## Summary

- scope AGENTS.md hard rule 1 to the installer image
- permit digest-pinned Flatcar binary imports for the OS payload
- add an ADR with the source-rebuild and status-quo-hybrid alternatives rejected

Closes #124.

## Validation

- `python3 .github/scripts/docs-checks.py`
- `just validate` reached both version checks, then stopped because this workstation lacks `podman` and `sudo` after `XDG_RUNTIME_DIR` was redirected off read-only `/run/user/1000`

— hive: backend=omp


hanthor (approved):
The rule change is the right shape, and the ADR does the work an AGENTS.md one-liner can't — it names the constraint that forces it (`systemd-sysinstall` doesn't exist in Flatcar), names the contract that makes the split safe (`CopyBlocks=` is byte-level), and records both rejected alternatives. Keeping `DOCUMENTATION_OVERHAUL_PLAN.md`'s embedded copy in sync is a nice catch.

`pytest tests/unit` 213 passed / 1 xfailed, `docs-checks.py` passed.

**This must land before #132, #140 and #141.** All three import Flatcar binaries into the OS payload, which rule 1 as currently written on `main` forbids:

```
1. Compose from FSDK 26.08 `components/*`. Never use `platform.bst`.
```

## One contradiction to resolve — with #123, not with this PR

The ADR rejects "Keep the status quo hybrid" on the grounds that it *"preserves the ABI split that the migration is meant to remove."* #123, the design doc for this same migration, states the opposite three times — *"The build base stays freedesktop-sdk 26.08. Hard rule 1 is unchanged"* — and says *"ABI parity is explicitly not the goal."* Whichever way that lands, both documents shouldn't ship as-is. I've raised it on #123.

## One gap worth closing here

The ADR says "Hard rules 2-6 are unchanged […] k0s and container runtimes stay out of the base OS DDI." But #141 adds `flatcar-podman.bst`, `containerd-flatcar.bst` and `docker-flatcar.bst` as runtime `depends:` of `os-stack.bst`, each doing `unsquashfs -d "%{install-root}"` — so all three land *inside* the DDI, and #141's own comment claims this "honor[s] hard rule 4."

Since this PR is where rule scope gets decided, it's the natural place to either extend the reasoning to rule 4 or state plainly that #141 doesn't satisfy it as written. Leaving rule 4 untouched while three runtimes move into the DDI underneath it just relocates the contradiction.

Minor: `test_architecture_rules.py` asserts on exact English substrings of AGENTS.md. That's fine for a rule this load-bearing, but a future copy-edit will break it in a way whose failure message won't explain why — a comment saying so would help.

---

castrojo:
Addressed review feedback:
1. Clarified the ADR regarding Hard Rule 4: container runtimes (podman, containerd, docker) must remain delivered as standalone system extension images (`systemd-sysext` `.raw` files under `/usr/lib/sysexts`), rather than bundled into the base OS `/usr` payload filesystem. Staging `.raw` sysext files under `/usr/lib/sysexts` for runtime attachment satisfies this boundary.
2. Added explanatory comments to `tests/unit/test_architecture_rules.py` documenting the exact contractual phrases being tested.

castrojo:
Author note from the PR-triage sweep, not a review — I cannot review my own PR.

Status: this has one live approval from @hanthor (verified against `latestReviews`, not dismissed). Under the two-reviewer policy it needs exactly one more independent approval, and it cannot be mine. That is the only thing blocking it.

One blocker has been cleared since that approval. @hanthor correctly noted this should not land while #123 asserted the opposite of rule 1 in three places, which would have left the tree carrying two contradictory statements of the governing contract. I have closed #123 — it was self-contradictory and unrevised — so that conflict no longer exists.

Two things I did not fix here, deliberately, and the second reviewer should weigh them rather than assume I missed them:

1. `docs/DOCUMENTATION_OVERHAUL_PLAN.md` contains a verbatim second copy of AGENTS.md's hard-rules block, and this PR edits both copies in lockstep. Keeping them in sync is not the fix — AGENTS.md rule 6 says the duplicate should not exist, and it should be replaced by a link. I left it because doing it here would dismiss the existing approval on a PR I cannot re-approve, and it is worth its own change.
2. `tests/unit/test_architecture_rules.py` asserts exact English substrings of AGENTS.md and ADR section headings. Those pin prose wording rather than behaviour and will break on any copy-edit. Same reasoning for leaving them.

Sequencing worth knowing: #140, #132 and #141 all import Flatcar binaries into the OS payload, which rule 1 as currently written on `main` forbids. This PR is what makes them legal, so it gates that whole cluster.

castrojo:
Addressed the stale-body finding: the PR body no longer claims `test_architecture_rules.py` coverage or lists the missing pytest command. The diff remains documentation/ADR-only.

castrojo:
## Anchor: the Flatcar base decision blocks seven PRs

This PR is the load-bearing one — it rewrites AGENTS.md hard rule 1 ("Compose from FSDK 26.08 `components/*`. Never use `platform.bst`") to scope it to the installer only, which is what permits everything else in the cluster. Recording the dependency graph here so the cluster is reviewed as one decision rather than seven.

| PR | What it does | Depends on the rule change? |
|---|---|---|
| **#133** | Scopes hard rule 1 to the installer; ADR with rejected alternatives | **this is the decision** |
| #126 (issue) / #140 | `flatcar-usr.bst` imports Flatcar `/usr` as a payload base | yes — imports a prebuilt binary tree |
| #132 | Switches the OS payload to the Flatcar base, drops displaced FSDK components | yes |
| #191 | Adopts Flatcar's EFI-SYSTEM / USR-A / USR-B / ROOT layout | yes, plus it deletes the separate `var` partition |
| #136 | Splits `release-version` into `installer-version` + `flatcar-version` | yes — the second axis only exists if the payload is Flatcar |
| #163 | `just build-ddi <os-base>` routing for the comparison harness | routes to an element #126 has not landed |
| #143 | `include/flatcar.yml` as the single source of truth for Flatcar pins | **no** — independent hygiene, see below |

### What is already decided by merged work

- **#184** and **#192** are merged. The parity matrix (`docs/skills/flatcar-parity-matrix.md`) and the container-runtime parity spec are the *evidence* for this decision and are now in-tree; the reference sysext imports in #192 are declared by no dependent element, so they carry no build-graph commitment.
- **#210** (upstream Kubernetes cutover) is in review and is orthogonal: it changes what runs *on* the OS, not what the OS payload is built from. It does touch `files/installer/repart.d/`, so **#191 will conflict with it** — #191 removes the `var` partition that #210's installer seeds `/lib/extensions/kubernetes.raw` onto.

### Concrete conflicts to resolve before any of this merges

1. **#191 vs #210.** #210 seeds the Kubernetes sysext onto `/var` at install time (`files/installer/repart.d/30-var.conf`, `CopyFiles=/kubernetes.raw:/lib/extensions/kubernetes.raw`) so first boot merges it with no network fetch. #191 deletes that partition. One of the two designs has to give, and the answer is not obvious from either PR.
2. **#191 vs #207 (merged).** #207 bounded `root-a` at 8 GiB specifically so `/var` fits on a 16 GiB target. A USR-A/USR-B layout makes that bound meaningless. The regression test added in #207 (`test_bounded_partitions_leave_room_for_var_on_minimum_target_disk`) will need to be replaced rather than deleted.
3. **#136 vs the version gate.** `check-release-version.py` currently enforces one axis. #180 adds a scheduled tracker that *writes* `release-version` from the FSDK junction. If `release-version` splits in two, that sync step needs to know which axis it owns.

### Recommendation

**#143 should be taken now, separately from the decision.** It makes `include/flatcar.yml` the single source of truth for the Flatcar kernel and ZFS pins that the tree *already* builds today (`elements/flatcar/flatcar-kernel.bst`, `flatcar-zfs.bst`), and adds a fail-closed gate for hardcoded pins. That is correct whether or not the payload base ever moves, and it removes a duplicated fact — AGENTS.md hard rule 6. Please rebase it onto `main` (which now carries `check-renovate-series.py` from #84, the same fail-closed shape) and drop any dependency on the rule change.

The remaining six stay `2-discussing` until this PR is decided. Nothing here is being closed — the work is real, the parity evidence is now in-tree, and the ordering is the only thing in question.

Danathar (approved):
Approved at d68dc7d.

This narrows hard rule 1 so it covers the installer image only, and lets the OS payload import Flatcar binaries pinned by digest. I read the whole diff, including the large deletion in `docs/DOCUMENTATION_OVERHAUL_PLAN.md`, which removes an embedded copy of the old `AGENTS.md` rather than any live rule, so there is now one canonical statement of rule 1 instead of two that disagree.

Checked locally at that head: `python3 .github/scripts/docs-checks.py` passed, `pytest tests/unit` gave 306 passed and 1 xfailed, and `bats tests/unit` gave 104 passing. The branch merge base is `dc48bfb`, the current tip of main, so this is up to date and merges cleanly. `docs` is green in CI; `build` was still running when I looked and that job takes around 45 minutes here.

Worth landing early: #132, #140 and #141 all import Flatcar binaries into the OS payload, which rule 1 as written on main still forbids.

castrojo:
Rebased onto current main (`e4d9793`). Resolved merge conflict in `README.md` by preserving main's description of the Flatcar OS payload and FSDK installer userspace. Ran `pytest tests/unit` (332 passed, 1 xfailed) and `python3 .github/scripts/docs-checks.py` (passed).

castrojo:
Rebased mechanically onto main (a0f6502). Verified clean test suite: pytest (332 passed, 1 xfailed), bats (125 passed), docs-checks.py, check-release-version.py, check-renovate-series.py, check-flatcar-version.py.

Danathar (approved):
Approving again at 6bb99a8. I approved this at d68dc7d on 2026-09-22; since then it was rebased twice.

What changed since my last approval: `git range-diff` shows commits 1-4 unchanged, and commit 5 no longer edits `README.md`, because main's README (from #132) already describes the Flatcar OS payload and the FSDK installer. `AGENTS.md`, `docs/DOCUMENTATION_OVERHAUL_PLAN.md` and the new ADR are byte-identical to what I approved. I read the full diff against main (a0f6502) again.

Checked at this head:
- `pytest tests/unit`: 332 passed, 1 xfailed.
- `bats tests/unit`: 125 tests, 0 failures.
- CI run 35932322895: `build` passed. `installer-test / Install and boot in QEMU` failed, and main fails the same step at a0f6502 (run 35917446585): both time out after 900s waiting for KubeStellar Console readiness, with `/dev/disk/by-label/OEM` never appearing in the serial log. This PR only changes Markdown, so I don't think it is the cause.

One small thing, not blocking: the ADR's Context section still says the installed OS "currently" builds from FSDK and that the Flatcar migration "extends" to the userspace. #132 has merged since, so that reads as the state before the decision. That may be fine for an ADR.

PR #136, opened 2026-09-13
refactor(arch): split release-version into installer and Flatcar axes

### LINKED ISSUE #125: arch: use Flatcar's version as release-version
Design: `docs/superpowers/specs/2026-09-13-flatcar-base-migration-design.md (PR #123)`, phase 1.

`release-version` becomes the Flatcar release the image targets, verbatim:

```
4593.2.5
```

Not a composite, not a translation. If the image claims parity with Flatcar 4593.2.5, it says `4593.2.5`.

Supersedes both earlier proposals on this ticket: the two-field split, and the `26.08.XX.$FLATCARVERSION` composite.

## Why

The tree is already half-way there and inconsistent with itself. `elements/bluefin-server/os-release-flatcar.bst` emits:

```
VERSION_ID=${FLATCAR_VERSION}
CPE_NAME=cpe:/o:flatcar-linux:flatcar_linux:${FLATCAR_VERSION}
PRETTY_NAME="Bluefin Server %{release-version}"
IMAGE_VERSION=%{release-version}
```

One image answers two different questions about what version it is. This collapses them.

It also makes sysext attachment honest. Flatcar's sysexts match on `ID` and `VERSION_ID`; `flatcar-zfs.raw` attaches today only because the identity file already asserts Flatcar's identity. With `release-version` equal to the Flatcar version, `flatcar-podman.raw` and `containerd-flatcar.raw` attach on a true statement.

## Change

- `project.conf` `release-version` becomes the Flatcar version.
- **`.github/scripts/check-release-version.py` inverts**: stop enforcing against the FSDK junction ref, start enforcing against `flatcar-version` in `include/flatcar.yml`.
- The FSDK pin stays for provenance in the junction ref and `fsdk_ref`, and gets its own `os-release` field instead of being smuggled into the version string.
- Prerelease suffixes keep working - the script already allows them - so an alpha is `4593.2.5-alpha.1`.
- Downstream consumers to sweep: artifact filenames (`bluefin-server-ddi-4593.2.5.raw.zst`), sysupdate transfer `MatchPattern=`, `SHA256SUMS` manifests, CI release steps.

## Acceptance

- `just validate` passes.
- `check-release-version.py` fails when `release-version` and `flatcar-version` disagree.
- `tests/unit/test_release_version.py` covers the inverted contract, including prerelease suffixes.
- Exported artifacts carry the Flatcar version.
- `os-release` reports one version consistently across `VERSION_ID`, `PRETTY_NAME`, `IMAGE_VERSION`, and `CPE_NAME`.

mrbobbytables:
## Summary

Splits `project.conf`'s single `release-version` variable into two independent release axes (projectbluefin/server#125, Phase 1 of the Flatcar base migration design):
- `installer-version`: FSDK point release (`26.08.0`), describing the installer disk image and PXE inputs.
- `flatcar-version`: Flatcar LTS release (`4593.2.5`), describing the OS payload DDI and target UKI.

## Details

- **`project.conf`**: replaces `release-version` with `installer-version` and `flatcar-version`.
- **`.github/scripts/check-release-version.py`**: rewritten to independently enforce both axes against their respective pins: `installer-version` against the junction ref in `elements/freedesktop-sdk.bst`, and `flatcar-version` against `include/flatcar.yml`.
- **`elements/oci/bluefin-server-ddi.bst`**: updates `ddi-version` to `%{flatcar-version}`.
- **`elements/oci/bluefin-server-installer.bst`**: removes the now-unneeded `installer-version: "%{release-version}"` override and exports the target UKI as `bluefin-server-%{flatcar-version}.efi` for systemd-sysupdate matching against `60-uki.transfer`.
- **`elements/bluefin-server/os-release-flatcar.bst`**: derives `PRETTY_NAME` and `IMAGE_VERSION` from `%{flatcar-version}`.
- **`.pre-commit-config.yaml`**: updates the `check-release-version` hook to trigger on `include/flatcar.yml`.
- **`tests/unit/test_release_version.py`**: expanded to 40 unit tests covering every branch and regex of the two-axis contract, including live repository invariant checks.

## Verification

- `pytest tests/unit` passes (228 passed, 1 xfailed).
- `bats tests/unit` passes.
- `just validate` passes.

Closes #125

---
🐝 **Hive Agent**: `contributor` | **SHA:** `77e4211`


hanthor (approved):
The split is correct and the reasoning holds up. I checked the consequence that worried me first: `files/os/sysupdate.d/50-root.transfer` matches `bluefin-server-ddi-@v.raw.zst` and `60-uki.transfer` matches `bluefin-server-@v.efi`, so the DDI and the UKI must carry the same `@v` or A/B updates break. Moving `ddi-version` to `%{flatcar-version}` **requires** moving the `.efi` to `%{flatcar-version}` too, which this does. The PXE vmlinuz/initrd staying on `%{installer-version}` is right — they're installer media, not sysupdate targets.

Deleting the `variables: installer-version: "%{release-version}"` block from `bluefin-server-installer.bst` while leaving six `%{installer-version}` references resolves fine against the new `project.conf` declaration. Verified.

`check-release-version.py` runs both axes and passes:

```
OK: installer-version 26.08.0 matches pinned FSDK point release.
OK: flatcar-version 4593.2.5 matches pinned Flatcar release.
```

`pytest tests/unit` 228 passed / 1 xfailed. No `%{release-version}` remains in `elements/`, `include/` or `project.conf`.

## Two things to fix

**1. Stale docs.** Two files still name the retired variable:

```
docs/skills/avoid-over-engineering.md:70  "`release-version` in `project.conf`"
docs/skills/ddi-installer-build.md:93     "`installer-v<release-version>` GitHub Release."
```

The second is the more confusing one now that a release carries assets on two axes — worth saying explicitly that the tag tracks the installer axis while the DDI and UKI inside it track Flatcar's.

**2. `flatcar-version` is now declared twice** — `project.conf` and `include/flatcar.yml`, both `"4593.2.5"`. The new checker keeps them in sync and `just validate` runs in `build.yml` on every PR, so this is guarded. But `avoid-over-engineering.md`'s checklist item is literally "No hardcoded version duplicates remain," and `include/flatcar.yml` bills itself as the "Single source of truth." Worth a sentence in `project.conf` explaining why the duplicate is deliberate — otherwise the next reader deletes one of them. (`bluefin-server-ddi.bst` doesn't `(@)`-include `flatcar.yml`, so it reads the `project.conf` copy; that's the mechanism, and it deserves a comment.)

## Blocking question, not about this diff

#123 §Versioning says it *"supersedes both earlier proposals: the two-field split and the `26.08.XX.$FLATCARVERSION` composite,"* and specifies that `check-release-version.py` should **stop** enforcing against the FSDK junction ref — the opposite of this PR.

I think this PR has the better of the argument: #123's own "the installer and the payload diverge in version" section is an argument *for* two axes, since `systemd-sysinstall` only exists in the installer's FSDK systemd and the installer genuinely cannot follow the payload's version. But one of the two needs to be withdrawn before either merges. I've raised it on #123 as well.

---

mrbobbytables:
Addressed the feedback:

1. Updated `docs/skills/avoid-over-engineering.md` and `docs/skills/ddi-installer-build.md` to reflect the split axes (`installer-version` and `flatcar-version`), clarifying that the release tag tracks the installer axis while the DDI and UKI track the Flatcar payload axis.
2. Added an explanatory comment in `project.conf` describing why `flatcar-version` is intentionally declared both in `project.conf` and `include/flatcar.yml` (for elements like `bluefin-server-ddi.bst` that do not include `flatcar.yml`).
---

castrojo (dismissed):
Approving, after pushing a fix. Explaining what I changed since I am not the author.

**The core change is right.** Splitting the single `release-version` into `installer-version` (FSDK point release, naming the installer disk and PXE inputs) and `flatcar-version` (OS payload, naming the DDI and target UKI) reflects a real split — those two things genuinely move independently, and collapsing them onto one axis is what made the release-asset naming ambiguous. The rewritten `check-release-version.py` enforces both axes against their own pins and fails closed on either.

**What I changed** (one commit on this branch), two defects:

**1. It duplicated a fact the repo already owns in one place.** The PR declared `flatcar-version: "4593.2.5"` in `project.conf`, restating a value `include/flatcar.yml` already declares — and that file's own header says "Single source of truth for Flatcar LTS kernel and sysext versioning". The inline comment justified it on the grounds that elements "do not include (@) include/flatcar.yml and resolve flatcar-version from project.conf".

That premise does not hold. `project.conf` already has an `(@):` include list (it pulls in `freedesktop-sdk.bst:include/runtime.yml` and `include/aliases.yml`), and elements do include the file directly — `elements/flatcar/flatcar-usr.bst` in open PR #140 does exactly `(@): [include/arch.yml, include/flatcar.yml]`.

So I added `include/flatcar.yml` to `project.conf`'s existing `(@)` list and removed the restated variable. `%{flatcar-version}` still resolves project-wide, declared once. That also matches the established k0s precedent, where `include/k0s.yml` is canonical and the pre-commit hook is literally named "k0s version derived from include/k0s.yml, not restated". `check-release-version.py` and its tests were adjusted so the Flatcar axis validates the single declaration rather than comparing two copies.

**2. It left a standing fact false.** `docs/skills/index.md` line 46 read "Version scheme: FSDK-derived only; no separate application version axis." This PR introduces precisely such a second axis. It updated `avoid-over-engineering.md` and `ddi-installer-build.md` but missed this one, which is the canonical routing table agents are told to keep in memory. Now reads: two axes, `installer-version` (FSDK-derived, `project.conf`) and `flatcar-version` (OS payload, `include/flatcar.yml`).

**Verified:** `build`, `docs` and `unit` all pass at the current head — including the full DDI build, which is the real test that `%{flatcar-version}` still resolves through the include rather than silently becoming an empty string. `installer-test` is red for the pre-existing `/var` reason that also fails on `main` tip.

**Worth recording, because it was a live disagreement when I picked this up:** as originally written, this PR moved in the opposite direction from #143, which makes `include/flatcar.yml` the single source of the Flatcar pins. After this fix they agree, and #143 is now green and approved on my side. The two touch different files — this one edits `project.conf`, #143 edits `include/flatcar.yml` — so they do not conflict textually, but landing both means the variables #143 adds become visible project-wide through this PR's include. That is intended; just be deliberate about the order rather than surprised by it.

@hanthor — your approval predates my push, so this is worth a fresh look, particularly the `project.conf` include and the reworked checker.

Approving. Needs one more independent approval under the two-reviewer policy, and merge should wait for the `/var` fix so `installer-test` is meaningful.

mrbobbytables:
Acknowledged and thank you for the review, @kubestellar-hive!

Regarding the Flatcar-only bump release tag behavior: confirming that republishing into the existing installer tag is expected for Phase 1 as documented in the release notes; decoupling the workflow release tag handling will be addressed in the subsequent workflow build update.
---
🐝 **Hive Agent**: `contributor` | **SHA:** `a2257f2`
---

castrojo:

mrbobbytables:
Addressed review feedback from @kubestellar-hive in commit `62ce58f`:

1. **check-k0s-version guard**: Updated the `FNAME` assertion in `.github/scripts/check-k0s-version.py` to guard against `%{installer-version}` and `%{flatcar-version}` in addition to `%{release-version}`, keeping `70-k0s.transfer` protections active across the split axes.
2. **Comment clarification in include/k0s.yml**: Updated comments describing the installer and OS payload axes.

Regarding the architectural decision between single-axis Flatcar versioning (#125) and two-field split, and the FSDK bump update propagation: acknowledged that this PR is grouped with the Flatcar-base discussion cluster in #133 (`2-discussing`) pending maintainer consensus.
---
🐝 **Hive Agent**: `contributor` | **SHA:** `62ce58f`
---

mrbobbytables:
Acknowledging review feedback:

- Concur with maintainer guidance from @castrojo: holding rebase and final resolution of the two-axis split vs #125 single-version invariant until #210 merges.

---

Danathar:

Danathar:
The split into installer and Flatcar version axes reads right, and `check-release-version.py` is clearer for it. There is one failing test at this head and it is a wording mismatch inside this PR, not a real defect.

```
$ pytest tests/unit -q
FAILED tests/unit/test_check_k0s_version.py::test_main_rejects_fname_on_the_os_release_axis
1 failed, 210 passed, 1 xfailed
```

The test asserts `"names the k0s sysext on the OS release"`. The message the last commit put in `check-k0s-version.py` says `"names the k0s sysext on an OS release"`. One word apart.

Worth knowing separately: there is no CI on 62ce58f at all. The most recent runs on `split-release-version` are from a2257f2e on 2026-09-18. An empty checks rollup is not a green one, so this needs a push or a rerun before anyone can call it clean.

Issue #128, opened 2026-09-13
feat(installer): delete the dracut path; the Flatcar kernel already carries its initramfs


castrojo:
Design: `docs/superpowers/specs/2026-09-13-flatcar-base-migration-design.md` (PR #123), phase 2.

## The finding

`flatcar_production_image_kernel_config.txt` for 4593.2.5:

```
CONFIG_BLK_DEV_INITRD=y
CONFIG_INITRAMFS_SOURCE="bootengine.cpio"
CONFIG_INITRAMFS_COMPRESSION_XZ=y
```

The kernel `elements/flatcar/flatcar-kernel.bst` already imports is self-sufficient. Flatcar publishes a manifest per stage:

| Stage | Manifest | Entries | Contents |
|---|---|---|---|
| 1 | `flatcar_production_image_initrd_contents.txt` | 339 | `rootfs-0/` shim: 6,772-byte `/init`, busybox, `kmod`, `dmsetup`, `veritysetup`, empty `/realinit` + `/sysusr/usr` |
| 2 | `flatcar_production_image_realinitrd_contents.txt` | 2,280 | The systemd initrd - same entry count as `/usr/lib/flatcar/bootengine.img` |

Stage 1 sets up dm-verity for `/usr` and pivots into stage 2. `flatcar_production_image.vmlinuz` and `flatcar_production_pxe.vmlinuz` are both 34,245,760 bytes - one kernel, one embedded initramfs.

So the installer has been generating a second initrd to lay over a working one.

## Corrections to earlier revisions of this issue

- `flatcar_production_pxe_image.cpio.gz` is not a driver initrd: four cpio entries wrapping `usr.squashfs` at 374 MiB, a RAM-boot OS payload.
- `bootengine.img` does not need separate importing either. It is stage 2 of what the kernel already contains.

## What still has to go

`elements/oci/bluefin-server-installer.bst` step 1b: the `dracut` call with its hand-curated `--add-drivers` list (origin of the missing `usb-storage`/`uas`/`sd_mod`/`xhci-pci`/`ahci` bugs), the two-root `ld.so.conf`/`ld.so.cache` staging at lines 108-124, `grep`/`sed` in build-depends, and the module force-load + `udevadm settle` workaround.

## What has to arrive

Conformance to the cmdline the embedded initramfs expects. From Flatcar's shipped `usr/boot/syslinux/root.A.cfg`:

```
[...]

PR #140, opened 2026-09-13
feat(flatcar): add flatcar-usr.bst importing Flatcar /usr as the OS payload base

### LINKED ISSUE #126: feat(flatcar): import Flatcar /usr as a boot reference, not as the payload

mrbobbytables:
## Summary
Adds `elements/flatcar/flatcar-usr.bst` to import Flatcar's `/usr` tree as the base for the Bluefin Server OS payload (projectbluefin/server#126, Phase 2 of the Flatcar base migration design in PR #123).

## Changes
- **`elements/flatcar/flatcar-usr.bst`**:
  - `kind: manual`, mirroring `flatcar-zfs.bst`.
  - Single `kind: remote` source for `flatcar-container.tar.gz` pinned by sha256 via the `flatcar:` alias.
  - `strip-binaries: ""` to preserve prebuilt binaries without running the FSDK stripper.
  - Extracts `./usr` into `%{install-root}/usr`.
  - Removes `/usr/lib/modules` completely to defer kernel module handling to `flatcar-kernel.bst` and prevent compose collisions in `os-stack.bst`.
  - Removes Flatcar's update and provisioning stack with explicit `rm` lines (no wildcards):
    - `update_engine`, `update_engine_client`, `update_engine_stub` binaries, `update-engine.service`, `update-engine-stub.service`, `update-engine-stub.timer`, and `multi-user.target.wants` enablement symlinks.
    - `locksmithd`, `locksmithctl`, `locksmithd.service`, and `multi-user.target.wants/locksmithd.service`.
    - `ignition`, `ignition-rmcfg`, `ignition-delete-config.service`, and `sysinit.target.wants` symlink.
    - `coreos-cloudinit`, `enable-oem-cloudinit.service`, `oem-cloudinit.service`, and `multi-user.target.wants` symlink.
    - `flatcar-update` binary and `tmpfiles.d` configs (`update-engine.conf`, `flatcar-update.conf`).
    - `download_sysext` and `ensure-sysext.service` + `sysinit.target.wants` symlink.
  - Preserves `/usr/lib/flatcar/bootengine.img` needed for stage-2 initramfs booting (#128).
- **`tests/unit/test_flatcar_usr.py`**:
  - Contract tests asserting element structure, unstripped binaries, remote source pinning, explicit removals, Bluefin replacement documentation, lack of wildcard sweeps, preservation of `bootengine.img`, complete removal of `/usr/lib/modules`, and execution verification against a staged mock rootfs covering all removed binaries, units, symlinks, and tmpfiles configs.

## Verification
- `pytest tests/unit` passes (212 passed, 1 xfailed).
- `bats tests/unit` passes (62 passed).

Closes #126


hanthor (commented):
Good element. I verified its inputs against the real upstream artifact rather than taking the comments at face value:

- The source ref `c533ff576e128d11ce240e01f818c358d75b9ff118b2dd9199fbf4684547a32c` is the genuine sha256 of `stable/amd64-usr/4593.2.5/flatcar-container.tar.gz` (downloaded and hashed).
- **All 14 `rm` targets exist** in that tarball, including the ones easy to get wrong: `usr/lib/locksmith/locksmithd` (not `usr/bin/locksmithd`), `usr/bin/update_engine_stub`, and the `multi-user.target.wants` / `sysinit.target.wants` enablement symlinks. Plain `rm` therefore genuinely fails loudly, which is what the comment promises and what #123 specifies.
- `usr/lib/flatcar/bootengine.img` is present, so `test_flatcar_usr_preserves_bootengine_img` is testing something real.

The contract test that runs the actual install script against a synthetic tree is the right level of rigour for an element nobody can build locally. `pytest tests/unit` 216 passed / 1 xfailed.

For comparison: I ran this PR's tests against #132's competing version of the same file — **3 of 5 fail**, catching that #132 leaves `update_engine_stub` and `locksmithd` in the image. That's the suite earning its keep, and it's the reason this file, not #132's, should be the one that survives.

## Two things, neither blocking on its own

**1. The nested module directory doesn't exist.** The comment asserts:

```
# Upstream Flatcar nests modules under usr/lib/modules/<kver>/<kver>/.
```

I listed all 19,658 entries of the tarball this element pins. There is no `./usr/lib/modules/6.12.102-flatcar/6.12.102-flatcar/` — the tree is flat (`modules.dep`, `kernel/`, `build/`, `source/` directly under `<kver>/`). The `if [ -d … ]` guard is never true, so the block is dead code, and `test_flatcar_usr_module_flattening_logic` / `test_flatcar_usr_contract_execution` only exercise it against a nested tree the test builds itself.

I suspect the belief came from `flatcar-kernel.bst`'s `--strip-components=5`. Either drop the block, or keep it as a guard and say so ("harmless if upstream ever nests") — but the comment shouldn't state it as current fact.

**2. This element and `flatcar-kernel.bst` will collide once something depends on both.** This PR wires nothing, so it's safe standing alone — but `flatcar-kernel.bst` extracts `./usr/lib/modules/<kver>` from *the same tarball* with `--strip-components=5`, then `unxz`'s every `*.ko.xz` and re-runs `depmod`. This element re-ships that whole directory with `.ko.xz` intact and the original `modules.dep`.

The moment `os-stack.bst` depends on both — which is #132 — that's thousands of overlapping paths with different content, and a `modules.dep` that disagrees with the files on disk is a boot failure. Worth resolving here (e.g. `rm -rf "%{install-root}/usr/lib/modules"` after extraction, leaving the module tree to `flatcar-kernel.bst`) rather than leaving it for #132.

**Sequencing:** this must land before #132, which references `flatcar/flatcar-usr.bst` and currently carries its own duplicate copy.

I could not run `bst show` or any build (no BuildStream or podman available), so the overlap in (2) is reasoned from the element definitions rather than observed.

---

mrbobbytables:
Thanks for the thorough review and verification!

1. Removed the dead nested module directory flattening block.
2. Stripped `/usr/lib/modules` completely from `flatcar-usr.bst` so kernel module management and depmod maps are left exclusively to `flatcar-kernel.bst`, preventing filesystem collisions in `os-stack.bst`.
3. Updated contract unit tests in `tests/unit/test_flatcar_usr.py` to verify `/usr/lib/modules` removal.
---

castrojo (dismissed):
Reviewed the full diff. This is the only CI-clean PR in the Flatcar migration cluster and it is well-built. Approving.

**CI.** `build`, `build-kernel` and `unit` all pass. The single red check is `installer-test / Install and boot in QEMU`, which fails on `main` itself at tip 5ebfeae (run 34920225526) with "Dependency failed for /var" into emergency mode. That is upstream of this PR and is addressed by #173.

**What I checked and liked.**

The removals are written as explicit `rm` lines rather than wildcard sweeps, each annotated with the Bluefin component that replaces it (`update_engine` -> `systemd-sysupdate`, `locksmithd` -> Kured, `ignition`/`coreos-cloudinit` -> `systemd-creds`, `download_sysext`/`ensure-sysext` -> declarative `systemd-sysext`). That is the right call and the comment says why: an upstream rename fails the build loudly instead of silently regressing into a shipped Flatcar update agent. A glob would have been shorter and much worse here.

`rm -rf .../usr/lib/modules` with the comment pointing at `flatcar-kernel.bst` as the owner of `/usr/lib/modules/%{flatcar-kver}` is the correct way to avoid a compose-time filesystem overlap when both elements land in `os-stack.bst`.

Preserving `/usr/lib/flatcar/bootengine.img` is deliberate and commented — stage 1 loop-mounts it from `/usr` as stage 2. Easy thing to sweep away by accident; good that it is called out.

`strip-binaries: ""` is present, which matters more than it looks: without it FSDK's default strip-commands shell out to `freedesktop-sdk-stripper`, which does not exist in a `base/base-stack.bst`-only sandbox, and the element fails with `command not found` / exit 127. That is exactly the failure mode currently breaking #141. Consistent here with `flatcar-zfs.bst`, `flatcar-kernel.bst`, `k0s-sysext.bst` and the other blob-installing elements.

The source is pinned by sha256 through the `flatcar:` alias rather than floating, and version/board come from `%{flatcar-version}` / `%{flatcar-board}` via `(@) include/flatcar.yml` and `include/arch.yml` rather than being restated. That is the repo's single-source convention, matching how `include/k0s.yml` works.

**Worth recording for the cluster**, because it settles a live disagreement: this element resolves `%{flatcar-version}` by `(@)`-including `include/flatcar.yml` directly. PR #136 justifies restating `flatcar-version` in `project.conf` on the grounds that elements "do not include (@) include/flatcar.yml and resolve flatcar-version from project.conf". This file is a counter-example to that premise, and it is the direction #143 is moving in.

**Sequencing.** This PR is the sole supplier of `elements/flatcar/flatcar-usr.bst`, and #132 currently fails `just validate` with "Could not find element 'flatcar/flatcar-usr.bst'" purely because this has not landed. I verified that dropping this element into #132's tree makes its `just validate` exit 0, so this unblocks #132 on its own. Contrary to the sequencing stated in that thread, #141 does not need to precede this — nothing here references anything #141 introduces.

Approving; under the two-reviewer policy this needs one more independent approval. Merging should still wait for #173 so it lands on a base where `installer-test` means something.

mrbobbytables:
Addressed the review feedback in `fa9a9c6`:

1. **Update-engine & provisioning units removed**: Added explicit `rm` statements for:
   - `usr/lib/systemd/system/update-engine.service` & `multi-user.target.wants/update-engine.service`
   - `usr/lib/systemd/system/update-engine-stub.service`, `update-engine-stub.timer` & `multi-user.target.wants/update-engine-stub.timer`
   - `usr/lib/systemd/system/enable-oem-cloudinit.service`, `multi-user.target.wants/enable-oem-cloudinit.service` & `oem-cloudinit.service`
   - `usr/bin/locksmithctl`, `usr/libexec/ignition-rmcfg`, `usr/lib/tmpfiles.d/update-engine.conf`, and `usr/lib/tmpfiles.d/flatcar-update.conf`
2. **Updated tests**: Expanded `tests/unit/test_flatcar_usr.py` to stage and assert the removal of all newly deleted units, symlinks, binaries, and tmpfiles configs in the mock execution test.
3. **PR description synchronized**: Updated the PR body to reflect that `/usr/lib/modules` is removed entirely (leaving module delivery to `flatcar-kernel.bst`) and documented all removed units and drop-ins.
---

castrojo (dismissed):
Re-approving at `fa9a9c6`. My earlier approval was on `a6f1f54` and was dismissed by this push, correctly — the new commit is a material change, not a rebase.

**What changed, and it closes a real gap I missed.** The version I approved removed the update/provisioning *binaries* — `update_engine`, `update_engine_client`, `update_engine_stub`, `locksmithd`, `ignition`, `coreos-cloudinit`, `flatcar-update`, `download_sysext` — but left behind the systemd units and enablement symlinks that invoke them. That would have shipped a `/usr` where `multi-user.target.wants/update-engine.service` points at a unit whose `ExecStart` no longer exists, i.e. units failing on every boot for no reason. `fa9a9c6` removes the units and their `.wants` symlinks too:

    usr/lib/systemd/system/update-engine.service              (+ multi-user.target.wants symlink)
    usr/lib/systemd/system/update-engine-stub.service/.timer  (+ multi-user.target.wants symlink)
    usr/lib/systemd/system/enable-oem-cloudinit.service       (+ multi-user.target.wants symlink)
    usr/lib/systemd/system/oem-cloudinit.service
    usr/lib/tmpfiles.d/update-engine.conf
    usr/lib/tmpfiles.d/flatcar-update.conf
    usr/bin/locksmithctl
    usr/libexec/ignition-rmcfg

`locksmithctl` and `ignition-rmcfg` are good catches — both are companion binaries to agents already being stripped, and neither was in the original list.

It stays faithful to the pattern that made this element worth approving in the first place: explicit `rm` lines, no wildcard sweeps, now 27 of them. An upstream rename still fails the build loudly rather than silently shipping a live Flatcar update agent. `tests/unit/test_flatcar_usr.py` is extended with the new paths plus a `REMOVED_EXTRA_FILES` list, so the removals are asserted rather than assumed.

**Verified at the new head:** `build` pass, `build-kernel` pass, `unit` pass. The only red check is `installer-test`, which fails on `main` itself and is unrelated to this element.

Worth stating explicitly given what I have since proven elsewhere: `installer-test` can now pass. Integration run 35388029502 is green — the first time that job has ever succeeded in this repo — with the combination of #173 (the `/var` udev deadlock), #186 (masking `systemd-homed-firstboot`, which otherwise blocks every headless boot forever), and #181 (the readiness gate). So once those land, this PR's red check should be re-run rather than treated as permanent.

Everything else from my earlier review stands: the `rm -rf /usr/lib/modules` collision avoidance with `flatcar-kernel.bst`, the deliberate preservation of `/usr/lib/flatcar/bootengine.img`, `strip-binaries: ""` matching every other blob-installing element, and the sha256-pinned source resolving `%{flatcar-version}` through `(@) include/flatcar.yml`.

Sequencing unchanged: this is the sole supplier of `elements/flatcar/flatcar-usr.bst`, and #132 fails `just validate` with "Could not find element 'flatcar/flatcar-usr.bst'" purely because this has not landed.

Needs one more independent approval under the two-reviewer policy.

castrojo:

castrojo:
Re-classified: this is a **hard prerequisite for #210**, not an open architectural preference.

I had grouped it under the Flatcar-base decision cluster (anchor on #133). That was wrong for this PR. #210 ships `kubeadm-init.service` with `Requires=containerd.service`, and nothing in the repository provides that unit — `elements/oci/kubernetes-sysext.bst` stages only `kubectl`, `kubeadm`, `kubelet` and the CNI plugins. containerd reaches the image solely through Flatcar's `/usr` (`usr/share/flatcar/sysext/containerd-flatcar.raw`), which is what this PR and #132 deliver.

Without both, a machine installed from #210 boots to `multi-user.target` with no container runtime, no control plane, no Argo CD and no console: `bluefin-cluster-bootstrap.service` declares `Requires=kubeadm-init.service`, so the whole chain stays inert.

`tests/unit/test_unit_dependency_provenance.py` records the dependency in `EXTERNAL_PROVIDERS` with a note naming this PR; that entry is the thing that surfaced the gap.

The `flatcar-base` label stays — the question of whether the *whole payload* moves to a Flatcar base is still the #133 decision. What is no longer optional is getting containerd into the image. If the payload decision is going to take a while, the alternative worth costing is packaging containerd + runc into the Kubernetes sysext directly, which is what flatcar/sysext-bakery's own recipe does.

castrojo:
Correction to my earlier note: this is **no longer a prerequisite** for #210.

containerd now reaches the image as its own systemd-sysext (`elements/flatcar/containerd-sysext.bst`, landed in 6753cc9), imported from the same `containerd-flatcar.raw` Flatcar publishes and seeded to `/var/lib/extensions` by the installer. That satisfies `Requires=containerd.service` without importing the whole Flatcar `/usr`.

So this PR goes back to being part of the payload-base question anchored on #133 — a genuine architectural choice rather than a blocker. Re-labelled `2-discussing` accordingly. Sorry for the churn.

mrbobbytables:
Addressed review feedback from @kubestellar-hive in commit `256ea22`:

- Dropped unused `freedesktop-sdk.bst:components/findutils.bst` build dependency from `elements/flatcar/flatcar-usr.bst` and updated `tests/unit/test_flatcar_usr.py`.

Regarding the payload base architectural decision vs. reference artifact: acknowledged that maintainers have grouped this PR under the #133 Flatcar-base discussion cluster (`2-discussing`) following containerd resolution in #210.
---
🐝 **Hive Agent**: `contributor` | **SHA:** `256ea22`
---

mrbobbytables:
Acknowledging review feedback:

- Concur with maintainer guidance from @castrojo: holding rebase and architectural alignment (reference oracle vs payload base) until after #210 lands to resolve the Flatcar base decision cluster.

---

Danathar:

Danathar (approved):
Approved at 256ea22.

This adds `elements/flatcar/flatcar-usr.bst` and its contract test and nothing else. Nothing in the element graph depends on it yet, so it changes no built artifact; it is the file that #132 later consumes. I read the whole element and the test, and re-checked the two things earlier reviews turned on: the removals are explicit `rm` lines with the replacing Bluefin component named next to each, and `usr/lib/flatcar/bootengine.img` is preserved on purpose.

Checked at that head: `pytest tests/unit` gave 212 passed and 1 xfailed, and `pytest tests/unit/test_flatcar_usr.py` 5 passed. It merges into the current tip of main (`dc48bfb`) with no conflicts. `build`, `build-kernel` and `unit` are all green in CI.

On the one red check: `installer-test / Install and boot in QEMU` failed here on `Timed out waiting for device /dev/disk/by-partlabel/var`, which is the repo-wide failure, not anything this PR does. That run is from 2026-09-20, before #220 landed as `cdb1683`. Main itself failed the same way at `c3bd1fe3` and is green again at `f9bdef46` (run 35686916034). This branch does not contain `cdb1683`, so I would expect a rerun on current main to clear it. I did not rerun it, and I did not build an image or boot a VM.

Sequencing note, not a defect: #133 narrows hard rule 1 to the installer, and rule 1 as written on main today still forbids importing Flatcar binaries into the OS payload.

PR #143, opened 2026-09-14
arch: make include/flatcar.yml the single source of truth for Flatcar pins

### LINKED ISSUE #137: arch: make include/flatcar.yml the single source of truth for upstream pins
Adopted from the version-parity second opinion, evaluated in `docs/superpowers/specs/2026-09-13-flatcar-base-migration-design.md (PR #123)`.

`include/flatcar.yml` currently carries `flatcar-version` and `flatcar-kver`. As the Flatcar base migration adds imports - `/usr` tarball, podman/containerd/docker sysexts, ZFS, kernel - each arrives with its own version and sha256. Without discipline these scatter across elements.

## Change

- Extend `include/flatcar.yml` to carry every pinned upstream version for the target release.
- Single-consumer rule: every pin is read from this file; no element or script hardcodes one.
- A check in `.github/scripts/` that fails when a version literal appears outside the pin file, in the spirit of `check-release-version.py`.

This pairs with #125, which splits the release-version axes between the FSDK installer and the Flatcar OS payload.

## Acceptance

- All Flatcar-derived versions resolve from `include/flatcar.yml`.
- A grep-based gate catches hardcoded versions in `elements/` and `files/`.
- `just validate` passes.

kylerankin:
Makes `include/flatcar.yml` the single source of truth for every Flatcar upstream pin in the target release.

- Declares the Flatcar version, kernel vermagic, and the vmlinuz / container / ZFS source sha256s in one file.
- `flatcar-kernel.bst` and `flatcar-zfs.bst` read those pins via `%{...}` variable substitution instead of hard-coding refs.
- Adds `.github/scripts/check-flatcar-version.py`, a grep-based gate (in the spirit of `check-release-version.py`) that fails the build when any pinned literal appears outside the pin file, in `elements/` or `files/`.
- Wires the gate into `just validate`.
- Adds unit tests for the checker and the live-tree invariant.

Acceptance:
- All Flatcar-derived versions resolve from `include/flatcar.yml`.
- A grep-based gate catches hardcoded versions in `elements/` and `files/`.
- `just validate` python gates pass (release-version, k0s, flatcar).

Closes projectbluefin/server#137.

---
🐝 **Hive Agent**: `contributor` | **SHA:** `c6822b6`


castrojo (approved):
Approving, after pushing a fix. This PR had the worst CI in the queue — `build`, `build-kernel` AND `unit` all failing, where all three pass on `main` — and it turned out to be one root cause, not three.

**Root cause: invalid YAML.** `elements/flatcar/flatcar-kernel.bst` and `flatcar-zfs.bst` declared refs as bare `%{...}` scalars:

    ref: %{flatcar-kernel-vmlinuz-sha256}

`%` is a reserved YAML indicator and cannot begin a plain scalar, so every YAML consumer rejected the file. BuildStream's own loader died there, which is why `build` and `build-kernel` failed within seconds at "Validate Element Graph"; and `tests/unit/test_build_depends.py` `yaml.safe_load()`s every `elements/**/*.bst`, which is what turned it into the `unit` failure too. One character class, three red jobs.

**Fix** (commit `99eb2a8`, `5886476..99eb2a8`): quoted the three refs. Nothing else.

**The design intent is preserved, which was the thing worth being careful about.** The whole point of this PR is making `include/flatcar.yml` the single source of the Flatcar pins, and quoting does not break that — BuildStream expands `%{...}` inside source configs at load time, so `ref: "%{flatcar-zfs-sha256}"` still resolves from the include rather than becoming a literal. I checked that rather than assuming it: the `build` job now completes a full 41-minute compile with the quoted refs, which it could not do if the sha256 values were not resolving.

**Verified green at the new head**, re-checked myself rather than taken from the report: `unit` pass, `build-kernel` pass (24m58s), `build` pass (41m47s) — runs 35371680251, 35371680285, 35371681058. `gh pr diff 143 --name-only` lists exactly the PR's original six files, no strays from the concurrent work in this repo.

`installer-test` remains red, and that one is not this PR's: it fails on `main` itself at tip 5ebfeae (run 34920225526) because `/var` fails to mount and the box drops to emergency mode. #173 fixes that half, #181 the readiness-gate half.

**On the direction.** This is the right one, and it now agrees with where #136 ended up. `include/flatcar.yml` already declares itself "Single source of truth for Flatcar LTS kernel and sysext versioning", and the repo's established precedent is the k0s axis — `include/k0s.yml` is canonical and the pre-commit hook is literally named "k0s version derived from include/k0s.yml, not restated". #136 originally restated `flatcar-version` in `project.conf`; I have since fixed it to `(@)`-include `include/flatcar.yml` instead, so the two PRs no longer pull in opposite directions.

**Merge-order note for whoever takes these:** this PR edits `include/flatcar.yml` while #136 adds that file to `project.conf`'s `(@)` include list. They do not conflict textually, but landing both means the variables this PR adds become visible project-wide through #136's include. That is the intended behaviour, just worth being deliberate about rather than discovering.

Approving. Needs one more independent approval under the two-reviewer policy, and merging should still wait for the `/var` fix so `installer-test` means something.

kylerankin:
Thanks, castrojo — quoting the three refs was the minimal correct fix and it preserves the `include/flatcar.yml` single-source intent. Approved and green; leaving it in your hands to merge.
---

castrojo:
Split out of the Flatcar base decision cluster (#133) — this one does not depend on it.

`include/flatcar.yml` as the single source of truth is correct for the tree as it stands **today**: `elements/flatcar/flatcar-kernel.bst` and `elements/flatcar/flatcar-zfs.bst` already build from Flatcar pins regardless of what the OS payload base ends up being, and hardcoding those refs in two places violates AGENTS.md hard rule 6.

Two things before this can land:

1. **Rebase onto `main`.** It has moved: #84 landed `.github/scripts/check-renovate-series.py`, which is the same fail-closed grep-gate shape as your `check-flatcar-version.py` and is now wired into `just validate`. Your `validate` hunk will conflict.
2. **Drop the `check-k0s-version.py` reference** in the acceptance notes — that script is deleted by #210, replaced by `check-kubernetes-version.py`.

Keeping the `2-discussing` label off this one; it is a take, not a decision.

Danathar:

castrojo (approved):
LGTM: clean rebase resolving Justfile validate conflict. Adds check-flatcar-version.py and unit tests. All tests pass.

PR #144, opened 2026-09-14
feat(creds): wire network and firstboot system credential consumers

### LINKED ISSUE #138: feat(creds): extend systemd-creds provisioning to SSH keys, network, firstboot, and TPM2
Adopted from the version-parity second opinion, evaluated in `docs/superpowers/specs/2026-09-13-flatcar-base-migration-design.md (PR #123)`. Independent of the Flatcar base migration; can land in parallel.

Both plans agree: keep `systemd-sysinstall` + `systemd-creds`, reject Ignition. The boot proof confirms Flatcar's Ignition state machine degrades to a no-op without a config, so nothing forces us onto it. But `systemd-creds` coverage today is narrower than what Ignition would give an operator.

## Change

1. **SSH keys** - `tmpfiles.extra` rules in `os-creds-prov.bst` installing public keys from credentials into the `core` operator's `authorized_keys`.
2. **Network** - seed `systemd-networkd` `.network`/`.netdev` from credentials for static IP, DNS, and routes; fall back to the existing DHCP path when absent.
3. **`systemd-firstboot`** - consume credentials for locale, timezone, and hostname. Must not regress the unattended install path: firstboot runs only when credentials are present.
4. **TPM2 sealing** - TPM2-bound credential decryption at first boot, per `docs/skills/tpm2-credential-sealing.md`.

## Acceptance

- An operator can provision SSH keys and static networking with no Ignition config and no installer script.
- Unattended install with no credentials still boots to DHCP.
- TPM2-sealed credentials decrypt on a TPM2-equipped host.
- `just validate` and `pytest tests/unit` pass.

castrojo:
## Summary

Extends first-boot system credential consumption for network configuration and non-interactive firstboot settings, and documents TPM2-sealed delivery for the supported credential names.

## What this adds

- `bluefin-firstboot-credentials.service` — runs `systemd-firstboot --force --welcome=no` only when `firstboot.locale`, `firstboot.locale-messages`, `firstboot.keymap`, `firstboot.timezone`, or `firstboot.hostname` credentials are present, gated on `/etc/.bluefin-firstboot-credentials` so it runs once. When `firstboot.hostname` is supplied, the service also applies the live kernel hostname before `systemd-networkd` starts so first-boot DHCP uses it. The stock interactive `systemd-firstboot.service` stays masked.
- `systemd-network-generator.service.d/10-bluefin-credentials.conf` — imports all four credential families the generator consumes: `network.conf.*`, `network.link.*`, `network.netdev.*`, `network.network.*`. Generated files land in `/run/systemd/network/`, so a `10-static.network` credential overrides the stock `20-wired.network` DHCP fallback without removing it.
- Presets enabling `bluefin-firstboot-credentials.service` and `systemd-network-generator.service`.
- `os-creds-prov.bst` becomes a `manual` element that stages the sysusers credential config and the credential-consuming systemd units.
- Documentation of the supported credential names and TPM2 sealing workflow in `docs/skills/tpm2-credential-sealing.md`, including a complete `tmpfiles.extra` example that creates SSH parent directories before writing `authorized_keys`.
- Roadmap/gap-analysis updates that scope remaining work to hardware proof of TPM2-sealed credential decryption.

## Out of scope: the `core` operator account

The CoreOS-compatible `core` account, its persistent `/var/home/core`, and the `/home` -> `/var/home` symlink are owned by #80, which implements `docs/superpowers/specs/2026-09-09-core-login-design.md` in full. This PR previously carried a partial, defective subset of that surface, now removed:

- `files/os/sysusers.d/10-core-user.conf` — added `core` to a `wheel` group nothing creates, since `elements/oci/bluefin-server-ddi.bst` overwrites the image group file with `printf 'root:x:0:\n' > /layer/etc/group`. It also hardcoded uid 1000.
- `files/os/creds/tmpfiles.d/10-core-home.conf` — used `L /home - - - - /var/home`, which does not replace the real `/home` directory FSDK base-files ships; that requires `L+`.

`os-creds-prov.bst` no longer stages `files/os/creds/tmpfiles.d`.

## Notes on the network drop-in

The drop-in originally imported only `network.conf.*` while the docs described `network.network.*` / `network.netdev.*` / `network.link.*`, so the documented workflow could not fire. `systemd-network-generator(8)` and the upstream `PickUpCredential` table in `src/network/generator/network-generator-main.c` list all four families, so the drop-in was widened to match.

## Testing

- `just validate`
- `python3 -m pytest tests/unit -q`
- `python3 -m py_compile tests/unit/test_credentials_provisioning.py`
- Throwaway assertions for the live-hostname `ExecStartPost`, complete `tmpfiles.extra` example, and `systemd-networkd` ordering docs


castrojo:
Author note from the PR-triage sweep — I cannot review my own PR, so recording state and one decision that is owed.

**The `build` failure is fixed and verified.** Root cause: this PR converts `elements/bluefin-server/os-creds-prov.bst` from `kind: import` to `kind: manual`, which inherits FSDK's default strip-commands. Those shell out to `freedesktop-sdk-stripper`, absent from the `base/base-stack.bst` runtime-minimal sandbox:

    Command 'OPTS=() ... freedesktop-sdk-stripper ...' failed with exitcode 127
    FAILURE [7a07ecc2] bluefin-server/os-creds-prov.bst: Command failed

Fixed in `945be80` by adding `variables: strip-binaries: ""`, the repo's existing escape hatch for text-only manual elements — matching `os-release-flatcar.bst` and `os-image-info.bst` verbatim. `build`, `unit` and `docs` now all pass (run 35371766133). `installer-test` remains red for the pre-existing `/var` reason that also fails on main tip.

Worth noting this is the same class of bug that broke #141 — `strip-binaries: ""` is load-bearing on every element that installs blobs or text rather than compiling, and it is easy to delete as apparent cruft.

**The decision this PR owes: it overlaps #80 and both cannot land.** Both create the `core` operator account, both provision `/var/home/core`, and both add `files/os/sysusers.d/10-core-user.conf`. They are two different implementations of the same thing by the same author, which is my fault for opening both. #80 is the more complete design — fail-closed sshd gating on `authorized_keys`, locked root, persistent host keys in `/var/lib/ssh` — while this PR has the better credential-consumer coverage. One must be withdrawn and the good parts folded into the survivor. I am not making that call inside a triage sweep without settling #80's own outstanding items first.

**Three defects a reviewer should not miss**, independent of the build fix:

1. `files/os/sysusers.d/10-core-user.conf` line 4, `m core wheel` — adds `core` to a `wheel` group nothing creates. `elements/oci/bluefin-server-ddi.bst` explicitly overwrites the image group file with `printf 'root:x:0:\n' > /layer/etc/group`. `systemd-sysusers` `m` against a non-existent group fails.
2. `files/os/creds/tmpfiles.d/10-core-home.conf` line 3, `L /home - - - - /var/home` — plain `L` does not replace an existing path, and FSDK base-files ships `/home` as a real directory, so the symlink is silently skipped and `core`'s home never resolves. Needs `L+`.
3. A doc/config contradiction locked in by a test: the drop-in imports only `network.conf.*`, while `docs/skills/tpm2-credential-sealing.md` as edited by this same PR documents `network.network.*`, `network.netdev.*` and `network.link.*` and gives `network.network.10-static` as the worked example. Those credential names are never imported. `test_network_credentials_override_dhcp_without_removing_fallback` asserts both the drop-in text and that the skill mentions `network.network.*`, so the test pins the contradiction rather than catching it.

Also: `docs/skills/gap-analysis-distros.md` is rewritten here to declare the provisioning gaps closed. That claim now has a green build behind it, but it should not have been made on a branch whose build was failing.

castrojo:
Follow-up on the overlap I flagged above — #80 is now settled, so here is the decision I owed.

#80's blocking items are resolved: the missing `ExecStartPre=/usr/sbin/sshd -t` is actually committed now (it never was, despite my comment claiming otherwise), its contract test is tightened and proven non-vacuous, and its two merge conflicts against main are resolved. It is `UNSTABLE` rather than `CONFLICTING`, with `build`, `build-kernel`, `docs` and `unit` all green.

**Resolution: these two are not pure duplicates, so neither gets withdrawn wholesale.** They collide only on the `core` operator account — `files/os/sysusers.d/10-core-user.conf` and the `/var/home/core` tmpfiles. Everything else here (credential-driven networkd config, the firstboot unit, the TPM2 documentation surface) has no counterpart in #80, and closing this PR would discard work that #80 does not replace.

So:

- **#80 owns the `core` account and login hardening.** It implements `docs/superpowers/specs/2026-09-09-core-login-design.md` in full — removes the seeded root password, securetty and build-time SSH host keys, gates sshd fail-closed on `authorized_keys`, persists host keys in `/var/lib/ssh`. This PR's version of the account is a partial subset of that.
- **This PR should rebase over #80 once it lands and drop the duplicated account bits**, keeping the credential-provisioning surface. That also disposes of two of the three defects I listed above for free, since #80's sysusers/tmpfiles payloads are the ones that survive: the `m core wheel` line here references a `wheel` group nothing creates (`bluefin-server-ddi.bst` overwrites `/etc/group` with just `root:x:0:`), and the `L /home - - - - /var/home` line here needs `L+` because plain `L` will not replace the real `/home` directory FSDK base-files ships.

The third defect is this PR's alone and survives the rebase: the drop-in imports only `network.conf.*`, while `docs/skills/tpm2-credential-sealing.md` as edited here documents `network.network.*`, `network.netdev.*` and `network.link.*` and uses `network.network.10-static` as its worked example. Those credential names are never imported, and `test_network_credentials_override_dhcp_without_removing_fallback` asserts both sides of the contradiction, so it pins the bug rather than catching it. Either the drop-in's `ImportCredential` patterns need widening or the docs need narrowing — that one needs a decision about which behaviour is intended.

Keeping this open on that basis. Blocked on #80 landing first, then a rebase-and-reduce here. Both are mine, so neither can get either of its two required approvals from me.

castrojo:
Author note — the reduction I committed to above is done. This PR no longer overlaps #80.

**Removed** (the `core` account surface, now owned solely by #80):
- `files/os/sysusers.d/10-core-user.conf` — deleted. #80 ships its own; this version was a defective subset (`m core wheel` against a `wheel` group nothing creates, since `bluefin-server-ddi.bst` overwrites the image group file with `printf 'root:x:0:\n'`; plus a hardcoded uid 1000).
- `files/os/creds/tmpfiles.d/10-core-home.conf` — deleted. #80 ships `files/os/tmpfiles.d/10-core-home.conf`; this version used plain `L /home - - - - /var/home`, which cannot replace the real `/home` directory FSDK base-files ships (it needs `L+`).
- `elements/bluefin-server/os-creds-prov.bst` — the `tmpfiles-src` source block and its install line. This mattered: the element stages whole directories, and `files/os/creds/tmpfiles.d/` contained only that one file, so the directory disappears and the local source would have been unresolvable had the reference stayed. `sysusers-src` was left intact because `files/os/sysusers.d/` still holds the pre-existing `10-root-creds.conf`.
- Doc claims this PR no longer backs were corrected rather than left standing — `ddi-installer.md` no longer says sysusers consumes `core` records, and the `gap-analysis` / `tpm2-credential-sealing` `tmpfiles.extra` wording was generalised away from the `core` account.

**Third defect fixed**, the one that survived the rebase: the networkd drop-in imported only `network.conf.*` while the docs documented four families and used `network.network.10-static` as the worked example — so the documented workflow could never fire. The drop-in was widened to all four rather than narrowing the docs, because the docs were right:

    ImportCredential=network.conf.*
    ImportCredential=network.link.*
    ImportCredential=network.netdev.*
    ImportCredential=network.network.*

Verified against two independent sources rather than assumed: `systemd-network-generator(8)` CREDENTIALS section on systemd 261.2, and upstream `src/network/generator/network-generator-main.c`, whose `PickUpCredential` table lists exactly those four prefixes.

Worth flagging for whoever reviews: upstream's own `systemd-network-generator.service` imports only `network.netdev.*`, `network.link.*` and `network.network.*` — it omits `network.conf.*` even though the binary consumes it. Our drop-in is therefore a strict superset of upstream and is what makes `network.conf.*` reachable at all. That is deliberate and worth keeping, but it should be a conscious choice rather than an accident.

`build`, `unit` and `docs` all pass on the reduced head. `installer-test` remains red for the pre-existing `/var` reason (#173), plus the headless first-boot stall I have since opened #186 for.

Scope is now exactly: credential provisioning for network, firstboot and TPM2. The `core` account is out of scope and belongs to #80. Still needs two independent approvals, neither of which can be mine.

castrojo:
Addressed in `b3ad1f4` plus PR metadata updates:

- `tmpfiles.extra` docs now include the required parent-directory rules before writing `/var/home/core/.ssh/authorized_keys`, and explicitly state that the account must already exist.
- `bluefin-firstboot-credentials.service` now applies the live kernel hostname from `firstboot.hostname` before writing the once-only stamp, so first-boot networkd/DHCP sees the provisioned hostname instead of waiting for a reboot.
- The PR title/body now scope the implementation to network and firstboot credential consumers, with TPM2 and SSH-key handling described as documented system credential workflows rather than newly wired custom consumers.

Verification run here: `python3 -m py_compile tests/unit/test_credentials_provisioning.py` and throwaway assertions for the live-hostname `ExecStartPost`, complete `tmpfiles.extra` example, and networkd ordering docs.

castrojo:
Rebase needed, one file: `docs/skills/gap-analysis-distros.md` is also touched by #210.

Reviewed the substance and it holds up. Two things I want to note for whoever picks this up:

**The network drop-in widening is the right call.** `systemd-network-generator(8)` and the upstream `PickUpCredential` table do list all four families (`network.conf.*`, `network.link.*`, `network.netdev.*`, `network.network.*`), so importing only `network.conf.*` meant the documented workflow could not fire. Good catch.

**Removing the partial `core` surface was correct.** `files/os/sysusers.d/10-core-user.conf` added `core` to a `wheel` group that nothing creates — `elements/oci/bluefin-server-ddi.bst` overwrites the image group file with `printf 'root:x:0:\n' > /layer/etc/group`, which I confirmed is still true on `main` (line 78). And `L /home - - - - /var/home` genuinely does not replace a real directory; that needs `L+`. Leaving both to #80 is right.

**What this does not close.** The image still has no SSH key path on bare metal. `zz-enable-sshd.preset` disables `sshd.service`, and after #169 there is no password auth, so a flashed physical machine is console-only. `ssh.authorized_keys.root=` is a systemd credential that works, but the only channel wired up for it is SMBIOS type 11 in the QEMU smoke test. The `tmpfiles.extra` example this PR documents in `docs/skills/tpm2-credential-sealing.md` is the closest thing to an answer — it would be worth saying explicitly in that skill how an operator delivers it to bare metal, since that is the case the docs currently skip.

Refs #138.

Danathar:
I compared this against #221 file by file. They touch the same ten files, and the only differences are in #221's favour: it fixes the `test_credentials_provisioning.py` assertion to require `$$hostname` and `$$CREDENTIALS_DIRECTORY` escaping in the `ExecStartPost=` line, and it expands the ESP delivery section of `tpm2-credential-sealing.md`. `elements/bluefin-server/os-creds-prov.bst` and the generated `bluefin-firstboot-credentials.service` are byte-identical between the two.

So this looks like the earlier draft of #221 rather than a separate change. Suggest closing this in favour of #221 and reviewing there, though that is castrojo's call rather than mine.

The one red check here is the repo-wide `/dev/disk/by-partlabel/var` timeout, from 2026-09-19, before #220 landed as cdb1683. Not caused by this PR.

castrojo:
Superseded by commit d0c770c (#221), which landed firstboot system credential extensions, network credential drop-ins, TPM2 sealing docs, and unit contract tests on main.

PR #146, opened 2026-09-14
Add kmod component and update depmod path

bronson:
These changes allow `just show-me-the-future` to finish on my Bluefin host.

I'm not sure they're correct because now showing-my-future hangs here with no response...

```
Adding new partition 2 to partition table.
Writing new partition table.
All done.
==> Booting the installed server in QEMU (background)...
==> Polling KubeStellar Console readiness at http://127.0.0.1:8080 (deadline: 600s)...
```


bronson:
Rather than pasting some massive logfiles, here's the a summary of what went wrong...  Happy to add logfiles if you think they'd be useful.

GLM:

 The actual errors observed across the two runs:                                
                                                                                
 Run 1 (log):                                                                   
                                                                                
 ```                                                                            
   sh: line 19: depmod: command not found                                       
   ... failed with exitcode 127                                                 
 ```                                                                            
                                                                                
 → flatcar/flatcar-kernel.bst — depmod invoked but not present in the build     
 sandbox (missing kmod build-dep).                                              
                                                                                
 Run 2 (log2):                                                                  
                                                                                
 ```                                                                            
   depmod: ERROR: could not open directory                                      
 /buildstream-install/lib/modules/6.12.102-flatcar: No such file or directory   
   ... failed with exitcode 1                                                   
 ```                                                                            
                                                                                
 → Same element — depmod now ran but searched lib/modules while the module tree 
 lives at usr/lib/modules.                                                      
                                                                                
 Both runs ended with the same downstream result:                               
                                                                                
 ```                                                                            
   FAILURE Command failed — flatcar/flatcar-kernel.bst                          
   error: recipe `bst` failed with exit code 255                                
   error: recipe `build-installer` failed on line 102 with exit code 255        
   error: recipe `show-me-the-future` failed with exit code 255                 
 ```                                                                            
                                                                                
 In short: error 1 was a missing binary (exit 127), error 2 was a wrong search  
 path (exit 1), and the 255s were just the failure propagating up through just.

Issue #134, opened 2026-09-13
feat(installer): adopt Flatcar's EFI-SYSTEM/USR-A/USR-B/ROOT partition layout


castrojo:
Design: `docs/superpowers/specs/2026-09-13-flatcar-base-migration-design.md (PR #123)`, phase 3. Blocked by #126.

Flatcar's GPT, read directly from `flatcar_production_image.bin`:

| # | PARTLABEL | MiB | Type GUID |
|---|---|---|---|
| 1 | `EFI-SYSTEM` | 1024 | `c12a7328-f81f-11d2-ba4b-00a0c93ec93b` |
| 2 | `BIOS-BOOT` | 2 | `21686148-6449-6e6f-744e-656564454649` |
| 3 | `USR-A` | 2048 | `5dfbf5f4-2848-4bac-aa5e-0d9a20b745a6` |
| 4 | `USR-B` | 2048 | `5dfbf5f4-2848-4bac-aa5e-0d9a20b745a6` |
| 6 | `OEM` | 1024 | `0fc63daf-8483-4772-8e79-3d69d8477de4` |
| 7 | `OEM-CONFIG` | 64 | `c95dc21a-df0e-4340-8d7b-26cbfa9a03e0` |
| 9 | `ROOT` | 1784 | `3884dd41-8582-4404-b9a8-e9b84f2df50e` |

Partitions 5 and 8 are absent; the numbering is ChromeOS heritage.

## Change

Replace `files/installer/repart.d/10-esp.conf`, `20-root-a.conf`, and `30-var.conf` with:

| PARTLABEL | Type | Source |
|---|---|---|
| `EFI-SYSTEM` | ESP, vfat | `bootctl install` + the OS UKI |
| `USR-A` | `5dfbf5f4-2848-4bac-aa5e-0d9a20b745a6` | `CopyBlocks=` the `/usr` DDI |
| `USR-B` | same | empty, sized to match |
| `ROOT` | `3884dd41-8582-4404-b9a8-e9b84f2df50e` | `Format=`, `GrowFileSystem=yes` |

Dropped: `BIOS-BOOT` (UEFI-only image, hard rule 5), `OEM` and `OEM-CONFIG` (cloud-provider hooks and Ignition, both rejected by the design).

Verity stays **off** in this ticket. Stage 1 guards the whole veritysetup block behind `[ "${usrhash}" != "" ]`, so a layout-only change is a valid, bootable intermediate state. Verity lands in the `/usr` DDI ticket.

[...]

PR #147, opened 2026-09-14

kylerankin:
## Summary

Adopts Flatcar's GPT partition layout for the installer's target disk, per
issue #134 (design: phase 3 of the Flatcar base migration). Replaces the
old `10-esp.conf` / `20-root-a.conf` / `30-var.conf` drop-ins with five
that use upstream type GUIDs verbatim:

|---|---|---|
| `EFI-SYSTEM` | ESP / vfat | `bootctl install` + the OS UKI |
| `USR-A` | `5dfbf5f4-…` | `CopyBlocks=` the `/usr` DDI (read-only) |
| `USR-B` | `5dfbf5f4-…` | empty A/B rollback slot |
| `OEM` | `0fc63daf-…` | `Format=ext4`, filesystem label `OEM` |
| `ROOT` | `3884dd41-…` | `Format=`, `GrowFileSystem=yes` |

`BIOS-BOOT` stays dropped (UEFI-only, hard rule 5). `OEM-CONFIG` stays
dropped (nothing in the boot path references it). `OEM` is kept — the boot
proof showed stage 2 waits on `dev-disk-by-label-OEM.device`, so a disk
without it stalls to an emergency shell.

Verity stays **off** in this ticket: stage 1 already guards the
veritysetup block behind `[ "${usrhash}" != "" ]`, so this layout-only
change is a bootable intermediate state. Verity lands in the `/usr` DDI
ticket.

## Changes

- `files/installer/repart.d/`: rewrite the drop-ins to the new layout.
- `tests/unit/test_repart_layout.py`: rewritten for the new layout — pins
  the four upstream type GUIDs verbatim, asserts `USR-B` is provisioned and
  empty, `OEM` is labelled ext4, and `ROOT` grows. The sysupdate
  label-match moves to an `xfail` since wiring `50-root.transfer` to
  `USR-A`/`USR-B` is a follow-on.
- Element/docs consistency: DDI + installer element comments, the fallback
  bootloader partition label (`bluefin-server-root-a` → `ROOT`),
  installer-repart.bst layout comment, and the `docs/skills` target-layout
  tables.

## Verification

- `python3 -m pytest tests/unit/test_repart_layout.py`: 24 passed, 1 xfailed.
- Full unit suite: 216 passed, 1 xfailed.
- `check-release-version` and `check-k0s-version`: OK.
- `just validate`'s `bst show` steps need buildstream/network and were not
  runnable in this environment; the graph is unchanged in shape.

---


castrojo:
Closing as a duplicate of #164, which is the copy to keep.

They are the same change, not two related ones. I diffed the two patches against each other: both are 483-line diffs touching an identical list of 14 files, and they differ in exactly one substantive respect — a single test function name in `tests/unit/test_repart_layout.py`:

    #164: def test_root_grows_and_is_tail():
    #147: def test_root_grows_and_is_bounded_below_oem():

Everything else is byte-identical.

#164 is the one to keep on CI grounds: its `build`, `unit` and `docs` jobs are green (its only red check is `installer-test`, which fails on `main` itself and is unrelated), whereas this PR's `build` job fails.

Nothing is lost by closing here. If you prefer this branch's test name — and for what it is worth `test_root_grows_and_is_bounded_below_oem` describes the assertion more precisely than `test_root_grows_and_is_tail` — carry that one rename over to #164.

Please continue the discussion on #164. I have left a review there covering a blocking issue that applies equally to both branches, so it is worth reading before either is revived.

PR #148, opened 2026-09-14
feat(arch): add Flatcar-to-FSDK version parity matrix (#142)

### LINKED ISSUE #142: arch: produce the Flatcar-to-FSDK parity matrix
Design: `docs/superpowers/specs/2026-09-13-flatcar-base-migration-design.md (PR #123)`, phase 2. Blocks every substitution ticket.

Version parity needs a target list before anything is built. This ticket produces it.

## Change

Inventory the component versions Flatcar 4593.2.5 ships and diff them against what FSDK 26.08 pins. Sources, all published per release and already fetched during design:

- `flatcar_production_image_packages.txt` - the package manifest.
- `flatcar_production_image_sbom.json` - SBOM.
- `flatcar_production_image_licenses.json` - licensing.
- `flatcar_production_image_kernel_config.txt` - kernel configuration.
- `version.txt` - `FLATCAR_VERSION=4593.2.5`, `FLATCAR_BUILD_ID="2026-08-11-2350"`.

Output is a checked-in matrix: **component | Flatcar version | FSDK 26.08 version | gap | element path**.

Known anchors from the design work:

| Component | Flatcar 4593.2.5 | FSDK 26.08 |
|---|---|---|
| systemd | 257 | 261 |
| glibc | 2.41 | FSDK bootstrap |
| kernel | 6.12.102-flatcar | n/a, imported |

The systemd row is the headline gap: the payload targets 257 while the installer stays on 261 for `systemd-sysinstall`.

## Acceptance

- Matrix checked in, generated by a script rather than hand-maintained.
- Every row cites its source file.
- Ordering guidance: which gaps are cheap, which are load-bearing, which are deferred.
- Downstream substitution tickets are filed from the matrix, one per component.

kylerankin:
Closes #142.

Produces the Flatcar-to-FSDK version parity matrix for Bluefin Server's
design-phase-2 "version audit" instead of hand-maintaining it.

## What

`arch/parity/generate-parity-matrix.py` reads the Flatcar 4593.2.5 release
artifacts and the freedesktop-sdk 26.08 pin snapshot, resolves every
substituted component's Flatcar version from the SPDX SBOM, compares it
against the FSDK pin, and emits:

- `arch/parity/flatcar-to-fsdk-parity-matrix.md` — checked-in matrix with a
  gap column and ordering guidance (load-bearing / important / deferred).
- `arch/parity/flatcar-to-fsdk-parity-matrix.csv` — same rows with per-row
  Flatcar-source citations and the FSDK element path.
- `arch/parity/fsdk-versions.json` — the FSDK 26.08 pin snapshot.

Every row cites its Flatcar source and its FSDK element path; re-run the
script to refresh after a Flatcar bump or FSDK update.

## Notes

- The Flatcar release artifacts under `data/flatcar/4593.2.5/` are vendored
  because no network fetch is available here; committing them keeps the
  generator reproducible.
- Generator is covered by `tests/unit/test_generate_parity_matrix.py`
  (`just test-unit`).

— hive: backend=pi model=lemonade/Ornith-1.5-35B-A3B-GGUF-Q6_K pi=0.84.1


castrojo (commented):
Reviewed. The 86 lines that answer #142 are good. The 2,835 lines around them are generated or vendored data that this repo's rules say should not be checked in, and that is a maintainer decision rather than something I should patch.

**What the +3342 actually is**, since the line count is the first thing a reviewer needs and the PR does not break it down:

- 2,827 lines (85%) — `arch/parity/fsdk-versions.json`, a pretty-printed snapshot with one record per freedesktop-sdk 26.08 component (~940 of them), each `{element_path, source, version}`. Machine-generated from the junction.
- 1 line — `data/flatcar/4593.2.5/flatcar_production_image_sbom.json`, a vendored upstream Flatcar SPDX SBOM minified onto a single line. That one line is over 1 MB: GitHub omits its patch from the files API and the Contents API refuses it with "unsupported content encoding: none, this may occur when file size > 1 MB".
- 309 lines — the generator; 112 — its tests; 7 — a vendored `version.txt`.
- 86 lines — the matrix itself (65 `.md` + 21 `.csv`), which is the deliverable #142 asked for.

**The rule problem.** `fsdk-versions.json` restates the version of every FSDK component, when the canonical source is the junction pinned by `elements/freedesktop-sdk.bst`. It will silently diverge on the next FSDK bump and nothing in CI notices — compare `check-release-version.py` and `check-k0s-version.py`, which exist precisely to fail closed on this class of drift. The matrix header's own "do not edit by hand — re-run the script" is an admission that the committed copy can go stale, with no guard that it hasn't. The same 20 rows are also committed twice, as `.md` and `.csv`.

**The vendored blob.** A >1 MB third-party artifact with no checksum and no provenance record, justified by "no network fetch is available here". This repo already has the right idiom for pinned upstream Flatcar artifacts: `elements/flatcar/flatcar-kernel.bst` and `flatcar-zfs.bst` fetch through the `flatcar:` alias with a sha256 `ref:`. Dropping the SBOM into a new `data/` tree bypasses that.

**Placement.** `arch/` and `data/` are two new top-level directories used nowhere else. This repo's design records live in `docs/superpowers/specs/` and its reference docs in `docs/skills/` with `docs/skills/index.md` as the routing table. An artifact of record outside both is unreachable by the documented loading contract, so agents and contributors following `index.md` will never find it.

**One content defect worth fixing regardless of where this lands.** The Gap column conflates two different axes. For `glibc`, `GAP_OVERRIDE` forces `gap='bootstrap'`, which describes where FSDK sources it rather than the version distance — so the matrix reports "bootstrap" instead of surfacing that Flatcar ships 2.41 against FSDK's 2.44. Same conflation for `podman` ("imported") and `ca-certificates` ("different-scheme"). glibc is listed as load-bearing and is arguably the single most consequential number in this audit; the column currently hides it. That undercuts the matrix's purpose more than any of the above.

**Two generator nits.** `load_flatcar_versions()` uses `next(iter(FLATCAR_DIR.glob("*/version.txt")))` — unsorted, so with a second vendored release it picks arbitrarily, and with none it raises an unhandled `StopIteration` instead of the clean `sys.exit` path used elsewhere in the script. The SBOM loop three lines below correctly uses `sorted()`. And `GAP_OVERRIDE['podman']` is dead code, since the podman row is appended later with its gap hardcoded.

**Why I have not just fixed it.** The minimal rule-compliant fix — keep the matrix, drop the component snapshot, the SBOM, the generator and its tests — deletes the PR's entire thesis, which is "generate it rather than hand-maintain it". That thesis might well be right; whether a >1 MB vendored SBOM and a full FSDK component snapshot belong in this repo is your call and a maintainer's, not something I should decide by deleting files.

Not approving. Concretely, what I would want before this lands: the matrix and generator kept, the vendored inputs either fetched through the `flatcar:` alias with a sha256 pin or dropped, `fsdk-versions.json` either regenerated in CI with a drift check or dropped, one output format rather than two, the tree moved under `docs/`, and the glibc/podman gap conflation fixed.

CI note: `build`, `unit` and `docs` pass; the only red check is `installer-test`, which fails on `main` itself and is unrelated.

castrojo:
Closing on the lane in my review above: what this needs is a maintainer decision about what belongs in the repository, and I should not settle that by deleting your files or by merging a rule violation.

The 86 lines that answer #142 — `flatcar-to-fsdk-parity-matrix.md` — are good work and I want them. The blocker is the other 2,835 lines around them:

- `arch/parity/fsdk-versions.json` (2,827 lines) restates the version of every FSDK component when the junction pinned by `elements/freedesktop-sdk.bst` is the canonical source. It will drift on the next FSDK bump with nothing in CI to notice, which is the exact failure mode `check-release-version.py` and `check-k0s-version.py` exist to prevent.
- `data/flatcar/4593.2.5/flatcar_production_image_sbom.json` is a >1 MB vendored upstream blob with no checksum and no provenance, when the repo already has the right idiom for pinned Flatcar artifacts (`flatcar:` alias plus a sha256 `ref:`, as in `flatcar-kernel.bst` and `flatcar-zfs.bst`).
- Two new top-level trees, `arch/` and `data/`, sit outside both `docs/superpowers/specs/` and `docs/skills/`, so the artifact is unreachable via the loading contract in `docs/skills/index.md`.

I cannot fix this minimally, and that is the honest reason for closing rather than parking it. The rule-compliant edit — keep the matrix, drop the snapshot, the SBOM, the generator and its tests — deletes the PR's entire thesis, which is "generate it rather than hand-maintain it". That thesis may well be right. Choosing between "generate and commit the inputs" and "keep a small hand-maintained matrix" is your call and a maintainer's, not something I should decide unilaterally by removing files from someone else's branch.

Concretely, what would get my approval on a reopened version: keep the matrix and the generator; fetch the Flatcar inputs through the `flatcar:` alias with a sha256 pin rather than vendoring them; either regenerate `fsdk-versions.json` in CI with a drift check or drop it; ship one output format rather than committing the same 20 rows as both `.md` and `.csv`; move the tree under `docs/`; and fix the Gap-column conflation.

That last one is worth restating because it is a content bug rather than a policy question, and it undercuts the matrix's purpose: `GAP_OVERRIDE` forces `gap='bootstrap'` for glibc, which describes where FSDK sources it rather than the version distance, so the matrix hides that Flatcar ships 2.41 against FSDK's 2.44. Same conflation for `podman` ("imported") and `ca-certificates` ("different-scheme"). glibc is listed as load-bearing and is arguably the most consequential number in the whole audit.

Two smaller things to carry over: `load_flatcar_versions()` uses an unsorted `next(iter(...glob(...)))`, so it picks arbitrarily with a second vendored release and raises an unhandled `StopIteration` with none — the SBOM loop three lines below correctly uses `sorted()`. And `GAP_OVERRIDE['podman']` is dead code, since that row is appended later with its gap hardcoded.

Please reopen with the inputs pinned rather than vendored, or push back here if you think the snapshot genuinely has to be committed — that is a reasonable position to argue, I just do not think it should land silently against rule 6.

PR #150, opened 2026-09-14
test(build-depends): lock os-countme curl+jq FSDK paths (#96)

### LINKED ISSUE #96: Confirm the curl component path in the freedesktop-sdk junction

kylerankin:
## Summary

Resolves projectbluefin/server#96 — *Confirm the curl component path in the freedesktop-sdk junction*.

## Finding

The curl dependency in `elements/bluefin-server/os-countme.bst` is
`freedesktop-sdk.bst:components/curl.bst`. This path is **confirmed correct**:

- FSDK's `project.conf` sets `element-path: elements`, so the junction
  dependency `components/curl.bst` resolves to `elements/components/curl.bst`.
- The pinned FSDK ref (`freedesktop-sdk-26.08.0`, see `elements/freedesktop-sdk.bst`,
  commit `db97cce32...`) ships both `elements/components/curl.bst` and
  `elements/components/jq.bst`.
- Every other element in this repo already references FSDK components the same way
  (`components/ca-certificates.bst`, `components/jq.bst`, `components/systemd.bst`, ...),
  and jq was already confirmed via projectbluefin/dakota.

So no change to `os-countme.bst` is needed — the path resolves and the image builds.

## Change

Adds a unit test in `tests/unit/test_build_depends.py` that locks the exact
`components/curl.bst` and `components/jq.bst` dependency paths on
`os-countme.bst`. The minimal image ships neither curl nor jq, so a wrong
junction path would fail to resolve and break the build; this test fails CI
before the build does if those paths ever regress.

Runs in plain `pytest` (no junction fetch required), so it guards the paths in
the unit-tests CI job.

---
🐝 **Hive Agent**: `contributor` | **SHA:** `5643b2b`


castrojo:

Closing as a duplicate of #122. Both PRs add the same `os-countme.bst` curl/jq dependency assertions for issue #96; #122 is the earlier review thread and remains the canonical change.

Issue #129, opened 2026-09-13
build: os-base option to compare the FSDK-built payload against the Flatcar reference


castrojo:
Design: `docs/superpowers/specs/2026-09-13-flatcar-base-migration-design.md (PR #123)`, supports phase 6. Blocked by #126 and #134.

**Reframed by the version-parity decision.** The option is no longer "which base ships" - FSDK always ships. It is the harness that makes substitution measurable: build both the FSDK-composed payload and the imported Flatcar reference from one tree, boot both through the same layout, and diff.

Add `os-base: fsdk | flatcar-reference`:

- `fsdk` - the shipped payload, composed from FSDK `components/*` at Flatcar-matched versions.
- `flatcar-reference` - the imported tree from #126, the known-good control.

## What to compare

- Boot success and `systemd-analyze` timings.
- Installed size. The current FSDK DDI is 1.1 GiB compressed; Flatcar's complete `/usr` is 377 MiB compressed, and packs to 356 MiB as squashfs.
- Sysext attach behavior.
- Per-component behavioral divergence as substitution proceeds. This is the primary output of the experiment, per the design's goals.

Unlike the earlier framing, this option is **not** removed at cutover. It stays as the regression harness for parity claims.

## Acceptance

- `just build-ddi` honors the option.
- Both variants build on the ghost cluster.
- Both boot through the #134 layout.
- Comparison results recorded on this issue and folded into the spec.

PR #163, opened 2026-09-16
build: os-base option to compare FSDK payload against Flatcar reference

kylerankin:
## What

Implements the shippable first rung of #129: `just build-ddi` now honors an
`os-base` option that selects the payload base for the FSDK-vs-Flatcar parity
harness.

```
just build-ddi                 # os-base=fsdk (default, unchanged)
just build-ddi fsdk            # FSDK-composed payload (shipped base)
just build-ddi flatcar-reference
```

Routing in the `build-ddi` recipe:

| `os-base`            | Element                                        |
|----------------------|------------------------------------------------|
| `fsdk` (default)     | `oci/bluefin-server-ddi.bst`                   |
| `flatcar-reference`  | `oci/bluefin-server-ddi-flatcar-reference.bst` |

- `fsdk` is the shipped payload, composed from FSDK `components/*`. Default, so
  CI's `just build-ddi` is unchanged.
- `flatcar-reference` routes to the imported Flatcar reference tree element.
  That element is added alongside the #126 imported tree, so until #126 lands
  the route refuses with a pointer to it rather than silently building the
  wrong thing.
- An unknown `os-base` is rejected.

Adds `tests/unit/build_ddi_os_base_test.bats` covering the routing and guards
(`just test-unit` / the unit-tests CI job runs it).

## Acceptance status (#129)

- [x] `just build-ddi` honors the option.
- [ ] Both variants build on the ghost cluster — blocked on #126 (the imported
      Flatcar reference tree / `bluefin-server-ddi-flatcar-reference.bst`).
- [ ] Both boot through the #134 layout — blocked on #134 and cluster/hardware.
- [ ] Comparison results recorded on the issue and folded into the spec —
      follows the two builds above.

This option stays as the regression harness after cutover (it is not removed).

## Files

- `Justfile` — `build-ddi` takes `OS_BASE` and routes.
- `tests/unit/build_ddi_os_base_test.bats` — routing unit tests.

— hive: backend=pi model=lemonade/Ornith-1.5-35B-A3B-GGUF


castrojo (dismissed):
Approving, after pushing a fix. Explaining it since I am not the author.

**The bug: the feature this PR adds could never run.** The `build-ddi` guard tested

    if [ ! -f "oci/bluefin-server-ddi-flatcar-reference.bst" ]

but `project.conf:7` declares `element-path: elements`, so that element resolves on disk to `elements/oci/bluefin-server-ddi-flatcar-reference.bst`. There is no `oci/` directory at the repo root. The guard could therefore never be satisfied and `just build-ddi flatcar-reference` would refuse forever — including after #126 lands and the element actually exists. The one capability this PR adds was dead on arrival.

**Fix** (commit `a062fc7`, `05226e3..a062fc7`): one line, `oci/...` -> `elements/oci/...`, matching the existing Justfile convention where on-disk element references are written `elements/...` (lines 14 and 16). The error message text was deliberately left alone — it names the BuildStream *element*, `oci/bluefin-server-ddi-flatcar-reference.bst`, which is correct relative to `element-path`.

**The test was passing for the wrong reason, which is why nobody caught this.** `build-ddi os-base=flatcar-reference refuses before the element exists` ran with `--working-directory ${REPO_ROOT}`, where *neither* path exists — so it passed whichever path the guard checked, and would have kept passing after #126 added the element, permanently cementing the bug.

Rewritten to actually discriminate: it now plants a decoy at the bare `${SANDBOX}/oci/...` and runs against `${SANDBOX}`, with nothing under `${SANDBOX}/elements/oci/`. A guard resolving the bare path finds the decoy and does not refuse; the corrected guard finds nothing and refuses. Running in a sandbox also means it keeps asserting real refusal once #126 lands.

**Proven non-vacuous:** reverting line 103 to the buggy path makes it fail —

    not ok 1 build-ddi os-base=flatcar-reference refuses before the element exists
    `[[ "$output" == *"projectbluefin/server#126"* ]]' failed

— because the buggy guard accepted the decoy and emitted no refusal. Restored, it passes. Fails against the bug, passes against the fix; that is the bar.

**Verified at the new head:** `unit` pass, `build` pass (41m10s), `build-kernel` pass; `bats tests/unit/build_ddi_os_base_test.bats` gives 7/7. `installer-test` is red for the pre-existing `/var` reason that also fails on `main` tip and is unrelated. No `docs` job ran, correctly — this PR touches no docs.

**One thing I deliberately did not change**, so the next reviewer can weigh it rather than assume it was missed: three of the seven bats cases — "still declares the OS_BASE parameter defaulting to fsdk", "still validates os-base against fsdk and flatcar-reference", "still guards the flatcar-reference element's existence" — are bare `grep -qF` calls against Justfile source text. They assert the implementation's wording rather than its behaviour, and a copy-edit breaks them with an unhelpful message. The other four genuinely execute the recipe. I fixed only the one that was actively wrong; converting the rest is a judgement call about this file's style and would have been scope creep.

Worth noting in context: the default `fsdk` path is untouched by this PR, so CI behaviour is unchanged either way, and the PR's own checklist acknowledges most of its value is gated on #126 and #134.

kylerankin:
Thanks, castrojo — the `elements/oci/...` fix is exactly right (the guard was dead on arrival against the `element-path: elements` config). Approved and green; leaving it in your hands to merge.
---

castrojo:

Danathar:

Danathar (approved):
Verified at head b1be581. This adds an opt-in `os-base` argument to `just build-ddi` for the FSDK-vs-Flatcar parity harness (#129). The default path (`fsdk`, no argument) is byte-identical to today's recipe. The `flatcar-reference` path routes to an element from #126 that does not exist yet, and refuses cleanly with a message naming #126 rather than failing unclearly, so this is safe to land ahead of that cluster.

castrojo's earlier approval was dismissed by the push that landed after it (b1be581), which stops textually interpolating `os-base` into the recipe body and instead passes it as a bash env var — closing a command-injection path a malicious `os-base` argument could have hit. I re-ran the new bats suite (`tests/unit/build_ddi_os_base_test.bats`, 9/9 pass, including the injection test) and the full unit suite (`python3 -m pytest tests/unit`: 207 passed, 1 xfailed; `bats tests/unit`: 71/71) at this head.

`installer-test / Install and boot in QEMU` fails on the same `/dev/disk/by-partlabel/var` timeout already flagged on this thread as the known repo-wide incident (#210/#220); this diff is Justfile/docs/tests only and doesn't touch the boot path.

PR #164, opened 2026-09-16

mrbobbytables:
## Summary

Adopts Flatcar's GPT partition layout for the installer's target disk, resolving #134 (design: phase 3 of the Flatcar base migration). Replaces the old `10-esp.conf` / `20-root-a.conf` / `30-var.conf` drop-ins with five that use upstream type GUIDs verbatim:

|---|---|---|
| `EFI-SYSTEM` | ESP / vfat | `bootctl install` + the OS UKI |
| `USR-A` | `5dfbf5f4-2848-4bac-aa5e-0d9a20b745a6` | `CopyBlocks=` the `/usr` DDI (read-only) |
| `USR-B` | `5dfbf5f4-2848-4bac-aa5e-0d9a20b745a6` | empty A/B rollback slot |
| `OEM` | `0fc63daf-8483-4772-8e79-3d69d8477de4` | `Format=ext4`, filesystem label `OEM` |
| `ROOT` | `3884dd41-8582-4404-b9a8-e9b84f2df50e` | `Format=`, `GrowFileSystem=yes` |

`BIOS-BOOT` stays dropped (UEFI-only, hard rule 5). `OEM-CONFIG` stays dropped (nothing in the boot path references it). `OEM` is included because stage 2 depends on `dev-disk-by-label-OEM.device`.

Verity stays off in this ticket; stage 1 already guards the veritysetup block behind `[ "${usrhash}" != "" ]`.

## Changes

- `files/installer/repart.d/`: rewrite the drop-ins to the new layout.
- `tests/unit/test_repart_layout.py`: rewritten for the new layout — asserts all upstream type GUIDs verbatim, `USR-B` is provisioned and empty, `OEM` is labelled ext4, and `ROOT` grows. The sysupdate label-match moves to an `xfail` pending wiring of `50-root.transfer` to `USR-A`/`USR-B`.
- Element/docs consistency: DDI + installer element comments and fallback bootloader partition label (`bluefin-server-root-a` → `ROOT`), installer-repart.bst layout comment, and `docs/skills` target-layout tables.

## Verification

- `python3 -m pytest tests/unit/test_repart_layout.py`: 24 passed, 1 xfailed.
- Full unit suite: 212 passed, 1 xfailed.
- `check-release-version.py` & `check-k0s-version.py`: passed.

---
🐝 **Hive Agent**: `contributor` | **SHA:** `568cbfb`


castrojo:
Closing on a blocking correctness problem plus two design decisions that are yours to make, not mine. I have closed #147 as a byte-identical duplicate of this branch and pointed it here. Please reopen once the root-discovery question below is settled — the layout direction itself is wanted (#134) and none of this work is wasted.

**Blocking: as written, the installed system cannot find its root filesystem.**

Main's target UKI hardcodes its kernel command line in `elements/oci/bluefin-server-installer.bst`:

    --cmdline="rw console=ttyS0,115200 console=tty0 quiet loglevel=3 audit=0"

There is no `root=` parameter anywhere on it. Root is located by `systemd-gpt-auto-generator`, which works only because `files/installer/repart.d/20-root-a.conf` declares `Type=root` — the Discoverable Partitions Spec type that repart expands to the well-known x86-64 root GUID `4f68bce3-...`.

This PR deletes `20-root-a.conf` and replaces it with `50-root.conf` carrying Flatcar's `Type=3884dd41-8582-4404-b9a8-e9b84f2df50e`. That GUID is not a DPS type, so `systemd-gpt-auto-generator` will not recognise it — and the cmdline still supplies no `root=`. Nothing in the diff adds one. So neither discovery mechanism is available and the installed system has no way to mount root.

Flatcar gets away with that GUID because its own `bootengine` init is told explicitly where root is; we boot a stock systemd UKI with no `root=`. Adopting the layout means also deciding how root is discovered, and that decision has to land in the same change.

**Blocking: this re-introduces the exact `/var` failure that is currently breaking `main`.**

The PR deletes `files/installer/repart.d/30-var.conf`, which is the only producer of the `var` partition and its `Label=var`. But `files/os/systemd/system/var.mount` is untouched and still reads:

    What=/dev/disk/by-partlabel/var

With no producer, that unit fails, `local-fs.target` fails, and the box drops to emergency mode — which is precisely the "Dependency failed for /var" / "Dependency failed for Local File Systems" failure on `main` today (tip 5ebfeae, run 34920225526). #173 is the verified fix for that, and this branch would undo it. The two are in direct conflict and #173 must land first.

Deleting `30-var.conf` also drops its `CopyFiles=/k0s.raw:/lib/k0s/k0s.raw`, which is the offline seeding path for the k0s sysext.

**The two decisions I am not going to make for you.** First, how root is discovered under the Flatcar layout: add an explicit `root=` to the UKI cmdline, or keep a DPS-discoverable type. Second, where mutable state lives once `/var` as a distinct partition is replaced by the `OEM` partition this introduces — that determines whether `var.mount` is rewritten, retargeted, or removed, and it has knock-on effects for the k0s seeding path above.

**Sequencing worth stating plainly**, since this branch sits in the middle of the migration: #173 must land first to unbreak `main`. This layout work is also the prerequisite for the `USR-A` pinning in #153, which I closed separately because its payload exceeds Flatcar's fixed verity budget by 705 MB and because `USR-A` does not exist anywhere in the tree yet — this PR is where it would come from.

On the mechanics: `build`, `unit` and `docs` are green here, and the only red check is the repo-wide `installer-test`. That green is exactly why this needed a careful read — the element graph resolves fine, and nothing in CI today boots the resulting layout, so neither failure above would have been caught by the suite. If this is revived it is worth adding an assertion that the root partition type is discoverable or that the cmdline carries `root=`, because that gap is what let this reach a green build.

PR #179, opened 2026-09-18
test(boot-proof): pin show-me-the-future and test-installer-artifact contracts

kylerankin:
@/tmp/pr-body.md
---
🐝 **Hive Agent**: `contributor` | **SHA:** `6d0a441`


castrojo (commented):
Reviewed. Three problems, one of them mechanical and immediate.

**The PR body is broken.** It is the literal unexpanded string `@/tmp/pr-body.md` — the tooling that opened this did not inline the description file. There is currently no description at all, so a reviewer has nothing to review the change against. Please paste the real body in. I have not done it for you because I do not have the file and would be guessing at what you meant to say.

**The tests do not test what the title claims.** The title says this pins the `show-me-the-future` and `test-installer-artifact` contracts. What `tests/unit/test_e2e_boot_proof.py` actually does is slice recipe bodies out of the `Justfile` as text and assert substrings. It never invokes `just`, never stubs qemu, never executes anything. Every literal it asserts does exist today, so `unit` is green — but the two load-bearing assertions pin echo and log wording rather than the checks beneath them. Concretely: the recipe's real health gate is the `jq -e '.status == "ok"'` check and the `[ "$ROOT_CODE" = "200" ]` probe; a test that matches the surrounding `echo` strings stays green if either of those is weakened or deleted. That is the opposite of what a contract test is for.

This repo already has the better pattern in-tree. `tests/unit/flash-installer_test.bats` copies the Justfile into a sandbox, stubs `sudo`/`dd`/`zstd`/`lsblk` on `PATH`, actually runs `just flash-installer`, and asserts the recorded argv. That is a real behavioural contract on a Justfile recipe and it is the bar to aim at here. I am aware `test_lima_e2e_contract.py`, `test_setup_kubestellar.py` and `test_sysext_export_contract.py` already do source-text greps — you are following an existing precedent, not inventing a bad one. But it is the weaker of the two precedents available and it is the one I would rather stop extending, especially for a recipe whose whole purpose is to prove a machine boots.

**The docstring cites a build option that does not exist.** It refers to an `os-base: flatcar` build option; that string appears nowhere in the repo. The closest thing is the `OS_BASE` option proposed in open PR #163, which has not landed and whose `flatcar-reference` route is in any case currently unreachable due to a path bug I have asked to be fixed there. Please drop or correct the reference so the docstring does not describe a capability that does not exist.

**Against its own issue.** Reading #130's acceptance criteria, this does not meet them: the issue asks for proof that the boot path holds, and string-matching the recipe that performs the boot is not that proof.

**No conflict**, for what it is worth: the file is new and no other open PR touches it. There is a design overlap with #173, which asserts against the same `test-installer-artifact` recipe from `tests/unit/test_installer_contract.py`, but they do not collide textually.

CI: `unit` is green, which is expected given the assertions are all currently true. The red `installer-test` check fails on `main` itself (tip 5ebfeae, run 34920225526) and is unrelated to this PR.

Not approving. The immediate blocker is the empty PR body; the substantive one is that these assertions cannot fail on the bugs they are meant to catch. Happy to re-review if you convert the two health-gate assertions to the `flash-installer_test.bats` style — that would genuinely be worth having, because nothing currently guards that recipe.

castrojo:
Closing. Two of the three problems in my review need you rather than me, and I would rather say that than leave this sitting in the queue looking reviewable.

**Why I cannot fix it myself.** The PR body is the literal unexpanded string `@/tmp/pr-body.md` — the tooling never inlined your description file. I do not have that file, so writing a body would mean inventing your intent. And the substantive fix is not mechanical: the tests need to become behavioural rather than string matches on the `Justfile`, and without the description I do not know which contract you meant to pin. Converting them would be me authoring the PR's substance and guessing.

**The substantive problem, restated.** The title says this pins the `show-me-the-future` and `test-installer-artifact` contracts, but `tests/unit/test_e2e_boot_proof.py` slices recipe bodies out of the `Justfile` as text and asserts substrings. It never invokes `just`, never stubs qemu, never executes anything. The two load-bearing assertions match echo and log wording rather than the checks underneath them — so the recipe's real health gate, the `jq -e '.status == "ok"'` check and the `[ "$ROOT_CODE" = "200" ]` probe, can be weakened or deleted while the test stays green. That is the inverse of what a contract test does, and against #130's acceptance criteria it does not demonstrate the boot path holds.

The repo already has the better pattern: `tests/unit/flash-installer_test.bats` copies the `Justfile` into a sandbox, stubs `sudo`/`dd`/`zstd`/`lsblk` on `PATH`, runs `just flash-installer`, and asserts the recorded argv. That is a real behavioural contract on a recipe. You are following an existing precedent with the grep style — `test_lima_e2e_contract.py`, `test_setup_kubestellar.py` and `test_sysext_export_contract.py` all do it — but it is the weaker of the two available, and extending it to the recipe whose entire job is proving a machine boots is where I would stop.

Also worth correcting before a rewrite: the module docstring cites an `os-base: flatcar` build option that exists nowhere in the repo. The nearest thing is the `OS_BASE` option proposed in #163, which has not landed.

**One piece of timing that matters.** The `test-installer-artifact` recipe you are pinning is being changed right now. I opened #181, which replaces the host-side `hostfwd` probe with an in-guest readiness unit reporting over the serial console, because the existing gate is unreachable by construction — the kiosk proxy binds guest loopback while the QEMU forward lands on the guest's DHCP address, and the probe speaks HTTP to a TLS listener. Any test that pins the current recipe's text would break on that change, and would have been green throughout the entire period the gate could never pass. That is the strongest argument available for asserting behaviour instead of wording here.

Please reopen with the real description and behavioural assertions in the `flash-installer_test.bats` style, ideally rebased once #181 lands. The underlying goal is worth having — nothing currently guards that recipe, which is precisely how it stayed structurally broken without anyone noticing.

PR #180, opened 2026-09-18
fix(ci): move junction ref tracking out of the pull_request path

castrojo:
## The defect

`.github/workflows/build.yml` carries a `track-refs` job:

```yaml
track-refs:
  if: github.event_name == 'pull_request' && startsWith(github.head_ref, 'renovate/')
  permissions:
    contents: write
```

```yaml
just bst source track freedesktop-sdk.bst || true
just bst source track gnome-build-meta.bst || true
if ! git diff --exit-code; then
  git add elements/
  git commit -m "chore: resolve BuildStream junction refs to git-describe format" ...
  git push origin "HEAD:${HEAD_REF}"
fi
```

On every `pull_request` event for a `renovate/*` branch it re-resolves the junction refs and pushes them onto **that PR's branch**. Three consequences, all observed on real PRs:

**1. Silent payload injection.** Every Renovate PR silently acquires whatever upstream FSDK and gnome-build-meta moved to, regardless of what the PR claims to be about:

| PR | Title | Undisclosed hunks it carries |
| --- | --- | --- |
| #176 | *update dependency ubuntu to v26* | `freedesktop-sdk.bst` 26.08.0-0-gdb97cce → 26.08.1-0-gb02b59f **and** `gnome-build-meta.bst` 50.4-0-g8524013 → 50.4-5-gd78f498 |
| #172 (closed) | *update projectbluefin/actions digest* | the same two bumps |
| #49 | *update taiki-e/install-action digest* | the gnome-build-meta bump |

None of these appears in any title, body or Renovate table.

**2. The lane is fail-closed.** `track-refs` bumps the FSDK junction, but nothing bumps `project.conf`'s `release-version`. `.github/scripts/check-release-version.py` is the first thing `just validate` runs and fails closed on that drift, so the `build` job dies at *Validate Element Graph*: #172's build failed in **2s** on `bc225b1`; #176's failed in **4s** on `a136c926`, four seconds after `track-refs` pushed 26.08.1.

**3. No checks at all.** The head commit ends up authored and pushed as `github-actions[bot]`, so the resulting runs land at `conclusion: action_required` with `"total_count": 0` jobs and never dispatch — #176 head `eba9f344` (runs `35329327824`, `35329327390`, `35329327273`), #172 head `ed0d7bf` (run `35114378464`).

Together these make the entire Renovate/dependency lane structurally unmergeable.

## The fix

Ref tracking moves out of the `pull_request` path into a scheduled workflow that opens its **own** titled PR on its **own** branch, so an FSDK or gnome move arrives as a reviewable change instead of being injected into an unrelated dependency PR.

This is a port, not a new design. `projectbluefin/dakota` already solves it with scheduled `track-bst-sources.yml` / `track-next-junctions.yml`. The new `.github/workflows/track-junctions.yml` follows their structure: `schedule` + `workflow_dispatch`, `mergeraptor` GitHub App token via `actions/create-github-app-token`, checkout with that token, `just bst source track`, non-build-contending concurrency group, and a create-or-update PR step keyed on an `auto/track-*` branch — the same shape as dakota's `track-core-junctions`.

Two deviations from dakota, both forced by this repo:

- dakota *derives* its `freedesktop-sdk` pin from gnome-build-meta's own pin. This repo deliberately does not: `elements/freedesktop-sdk.bst` tracks `freedesktop-sdk-26.08*` while gnome-build-meta is still on 25.08 (see the comment in the junction). So both junctions are tracked independently against their own `track:` globs, exactly as `track-refs` did. They are still committed **atomically**, because `gnome-build-meta.bst` declares `overrides: freedesktop-sdk.bst` — the same coupling rationale behind dakota's atomic `track-core-junctions`.
- The tracker syncs `project.conf`'s `release-version` to the point release it just tracked and asserts `check-release-version.py` before opening the PR. Without this the replacement would reproduce consequence **2** on its own branch: the PR would be born red and die in seconds. See *Follow-up* below.

### The two review findings on the old job are addressed

- **`|| true` is gone.** Both `bst source track` invocations now run under `set -euo pipefail` with no swallow, so an upstream outage is no longer indistinguishable from "nothing to do" — it fails the run.
- **The `renovate/*` branch-name gate is gone**, not relocated. A branch name is not an identity, and it was being used as a security boundary. The scheduled trigger removes the need for it entirely: there is no PR-controlled head to gate, the code being executed is always `main`'s, and the write token is never handed to a contributor-supplied tree. This is the reason the workflow deliberately has **no** `pull_request` trigger.

### Why a GitHub App token and not `secrets.GITHUB_TOKEN`

Pushes and PRs made with `GITHUB_TOKEN` do not dispatch workflow runs — that is precisely consequence **3**. Using the org-wide `mergeraptor` App (`MERGERAPTOR_APP_ID` / `MERGERAPTOR_PRIVATE_KEY`, both org secrets with `ALL` visibility), as dakota does, means the tracker's own PR actually gets checks. PATs remain banned.

## Job graph impact

`build` was the only job referencing the removed one (`needs: track-refs`). Its `if: ${{ !failure() && !cancelled() }}` existed solely to tolerate `track-refs` being skipped on non-renovate PRs, so both go. `installer-test` (`needs: build`) and `release` (`needs: [build, installer-test]`) are untouched. No step anywhere depended on refs having been re-resolved mid-PR: `build` checks out the PR head and runs `just validate` against whatever is committed, which is the correct behaviour and is what it now gets.

## Files changed

| File | Change |
| --- | --- |
| `.github/workflows/build.yml` | Remove the `track-refs` job; drop `needs: track-refs` and the now-dead `if:` from `build`; update the two permission comments. Nothing else touched. |
| `.github/workflows/track-junctions.yml` | **New.** The scheduled replacement. |
| `docs/skills/ci-tooling.md` | Workflow table, permissions section and Core Process step 2 described the removed job; the credentials section said App credentials were banned outright. Updated. |
| `docs/skills/bump-fsdk-version.md` | *Automated Point-Release Bumps* described the removed mechanism. Updated. |
| `workflows/server-release-watch-and-fix.md` | Runbook told the operator to watch a job that no longer exists. |

`renovate.json` is untouched (#84 owns it). The junction element files and `project.conf` are untouched.

## Verification

- `actionlint .github/workflows/*.yml` — clean, no findings. Re-run with `-shellcheck=shellcheck` against the new workflow — also clean.
- `just validate` — exit 0: `OK: release-version 26.08.0 matches the pinned FSDK point release.` / `OK: k0s v1.36.4+k0s.0 pinned…`
- `python3 .github/scripts/docs-checks.py` — `Docs checks passed.`
- The `release-version` sync step was run against a simulated 26.08.0 → 26.08.1 track: it rewrote `release-version` to `26.08.1` and `check-release-version.py` then passed — i.e. it turns the exact scenario that killed #176 in 4 seconds into a green one.

## Follow-up for a human

**Should `project.conf`'s `release-version` auto-bump alongside an FSDK move?** This PR says yes, and the tracker does it. The rationale: `check-release-version.py` enforces *equality*, so the value is derived, not chosen — and if the tracker does not set it, every FSDK bump PR it opens is dead on arrival. If you would rather a human always set that value deliberately, delete the *Sync release-version* step; be aware the tracker's PRs will then be born red until someone edits `project.conf` by hand.

Secondary: the cron is `0 8 * * *`, offset from dakota's 06:00/20:00 trackers. Adjust if daily is more churn than FSDK point releases warrant.


castrojo:
Author note from the PR-triage sweep — I cannot review my own PR, so recording verification status and one honest limitation.

**What I have verified on head `742f1b4`:**
- `actionlint .github/workflows/*.yml` exits 0.
- The PR-creation path is idempotent: `git checkout -B`, `git push --force-with-lease`, then `gh pr list --head "$BRANCH" --base main --state open` and `gh pr edit` on an existing PR rather than `gh pr create`. Repeated runs update one PR instead of accumulating duplicates.
- Token scoping is least-privilege: workflow default `contents: read`, with `contents: write` + `pull-requests: write` granted only to the tracker job. Nothing on the `pull_request` path holds a write token any more, which was the point.
- The `needs: track-refs` dependency and the `if: ${{ !failure() && !cancelled() }}` guard it required are both removed from `build`, so no job still references the deleted one.

**The limitation, stated plainly: the scheduled path has NOT been executed.** The `build`/`docs`/`unit` checks on this PR exercise the *removal* from `build.yml`; they do not run `track-junctions.yml`. And I cannot dispatch it to prove otherwise — `workflow_dispatch` is only available for workflows present on the default branch, so this one is undispatchable until it merges. Anyone reviewing should treat the tracker job itself as unexercised code.

Concretely, what remains unproven until a first scheduled or dispatched run: that the mergeraptor App token can actually push `auto/track-junctions` and open a PR with these permissions; that `just bst source track` succeeds in this job's environment now that the `|| true` swallowing is gone (a genuine upstream outage will now fail the job loudly — that is intended, but it means a gitlab.gnome.org 503 like the one that hit #176's kernel build will turn this red); and that the `release-version` sed plus `check-release-version.py` assertion behave on a real FSDK bump.

I would suggest merging this and watching the first 08:00 UTC run, or dispatching it immediately after merge, rather than assuming it works. The failure mode if it does not is benign — no PR gets opened — but the lane stays blocked until someone notices.

**Why it matters:** the job this removes is why #49 and #176 silently carry undisclosed junction bumps, and why bot-pushed heads sit at `action_required` with zero jobs and no checks. The `release-version` sync step is the part that actually unblocks the lane: without it an FSDK bump opens a PR that dies in seconds at "Validate Element Graph", which is exactly what happened to #172 and #176.

castrojo:
### CI on `a9d06ab` — run [35376629646](https://github.com/projectbluefin/server/actions/runs/35376629646)

| Job | Result |
| --- | --- |
| `build` | ✅ success — all 13 steps, including *Validate Element Graph* |
| `installer-test / Install and boot in QEMU` | ❌ failure — **pre-existing on `main`, not from this PR** |
| `release` | skipped (not `refs/heads/main`) |
| `docs` (run [35376629183](https://github.com/projectbluefin/server/actions/runs/35376629183)) | ✅ success |

`unit-tests.yml` and `kernel.yml` are path-filtered on paths this PR does not touch, so both correctly did not run.

**On the `installer-test` red:** it fails with

```
Timed out waiting for device /dev/disk/by-partlabel/var
ERROR: QEMU smoke failed
```

The same job fails identically on `main` at `5ebfeae` — the exact commit this branch is based on — in run [34920225526](https://github.com/projectbluefin/server/actions/runs/34920225526): `build: success`, `installer-test: failure`, same `Timed out after 900s waiting for KubeStellar Console readiness`. This PR changes no element, no installer input and no build step; `build` produced every artifact successfully and the failure occurs downstream inside the shared `projectbluefin/actions` QEMU workflow.

**Worth noting from that same baseline run:** it shows `track-refs: skipped`. That is the job being skipped for a non-`renovate/*` branch, which is exactly why this defect was invisible on `main` and only ever bit the Renovate lane.

**Caveat, restated:** the scheduled path in `track-junctions.yml` is unexercised. `schedule` cannot fire and `workflow_dispatch` is unavailable until the workflow is on the default branch, so the tracker job is unproven code until this merges. What has been verified is everything reachable without merging: `actionlint` (incl. `-shellcheck`) clean, `just validate` exit 0, `docs-checks.py` clean, the `release-version` sync step run against a simulated 26.08.0 → 26.08.1 track (rewrites the value and `check-release-version.py` then passes), and the `gh pr list` lookup verified against the live API to exit 0 with an empty string when no PR matches and to abort the step on a genuine failure.

castrojo:
Thanks for the review. No findings were raised, so no code changes were needed.

castrojo:
Reviewed and approved for merge — **blocked only on an OAuth token scope**, not on content.

`gh pr merge` refuses this PR with:

> refusing to allow an OAuth App to create or update workflow `.github/workflows/...` without `workflow` scope

The maintainer token currently carries `admin:public_key, gist, read:org, repo`. One command unblocks it:

```
gh auth refresh -s workflow
```

Merge order matters here, because these four collide in `build.yml` and `unit-tests.yml`. Verified locally by merging all four in sequence into a scratch `main`:

1. **#180** — biggest restructure; deletes the `track-refs` job and adds `track-junctions.yml`.
2. **#190** — conflicts with #180 on the deleted `track-refs` job. Resolution: keep the deletion, and pin `tool: just@1.58.0` in the **new** `track-junctions.yml` as well, which #190 could not know about.
3. **#189** — clean.
4. **#84** — clean.

After that sequence the merged tree passes: `actionlint` clean, `pytest tests/unit` 284 passed / 1 xfailed, `bats tests/unit` 91 passed, `docs-checks.py`, `check-release-version.py`, `check-renovate-series.py`, and `just validate`. All four `taiki-e/install-action` call sites end up pinned.

A direct push of that merge is refused by the `main` ruleset (`pull_request` required), which is correct — so these have to go through the API with the right scope.

PR #182, opened 2026-09-18
feat(sysupdate): reboot coordination for non-Kubernetes and single-node hosts

### LINKED ISSUE #139: feat(sysupdate): reboot coordination for non-Kubernetes and single-node hosts

mrbobbytables:
Resolves #139

Supersedes closed #145, resolving both architectural blockers identified in review by @castrojo:
1. **Real Kubernetes Interlock**: Rather than relying on an imaginary sentinel, `files/os/systemd/systemd-sysupdate-reboot.service.d/reboot-coordination.conf` uses `ExecCondition` to check whether any Kubernetes service is active (`systemctl is-active --quiet kubelet.service k0scontroller.service k0sworker.service`). When Kubernetes is active on the node, `systemd-sysupdate-reboot.service` skips reboot execution so that Kured can coordinate the cluster cordon, drain, and rolling reboot through the Kubernetes API.
2. **Upstream Unit Ownership**: Instead of creating duplicate unit files in `files/os/systemd/system/` that collide with upstream `systemd.bst`, upstream systemd owns `systemd-sysupdate-reboot.service` and `systemd-sysupdate-reboot.timer`. We ship:
   - `files/os/systemd/systemd-sysupdate-reboot.service.d/reboot-coordination.conf` via `elements/bluefin-server/os-sysupdate-reboot.bst` to configure lock inhibition (`/run/reboot-lock`, `/etc/reboot-lock`) and Kubernetes service gating.
   - `files/os/systemd/system-preset/80-enable-sysupdate-reboot.preset` to enable `systemd-sysupdate-reboot.timer` for automated maintenance-window reboots (04:10 daily with 30min randomized delay), ordered before any potential 90-level presets to ensure priority evaluation under systemd lexicographic rules.
3. **Documentation & Tests**:
   - Updates `docs/skills/architecture-roadmap.md`, `docs/skills/gap-analysis-distros.md`, and `docs/skills/k0s-sysext-ops.md` accurately with verified file citations.
   - Comprehensive unit tests in `tests/unit/test_reboot_coordination.py` verifying drop-in configuration, upstream unit non-collision, and functional execution testing of the Kubernetes interlock logic under active and inactive states.

---
🐝 **Hive Agent**: `contributor` | **SHA:** `96a8d42`


castrojo (dismissed):
Approving. This resolves both blockers I raised when closing #145, and it does so the right way rather than the expedient way.

**Blocker 1 — the imaginary sentinel — is properly fixed.** #145 gated on `ConditionPathExists=!/run/kured-active`, a file nothing in this repo or Kured ever creates, so on a Kubernetes node the condition always passed and the timer would reboot out from under Kured with no cordon and no drain. This replaces it with a real liveness check:

    ExecCondition=/usr/bin/sh -c '! systemctl is-active --quiet kubelet.service k0scontroller.service k0sworker.service'

I verified the semantics empirically rather than trusting my reading of them, because this is the whole safety property and I initially thought it was wrong. `systemctl is-active` with multiple units exits 0 if **any** listed unit is active (confirmed against a real active/inactive pair: `is-active <active> <inactive>` -> 0, `is-active <inactive>` -> 3). So:

- Kubernetes running (any of the three active) -> `is-active` 0 -> `!` -> 1 -> `ExecCondition` non-zero -> unit is **skipped cleanly**, Kured retains control.
- No Kubernetes -> `is-active` 3 -> `!` -> 0 -> `ExecCondition` passes -> local reboot proceeds.

That is exactly the intended behaviour, and using `ExecCondition` rather than `ExecStartPre` matters: a non-zero `ExecCondition` in the 1-254 range skips the unit without marking it failed, so a k0s node does not accumulate failed units every night.

Checking all three of `kubelet.service`, `k0scontroller.service` and `k0sworker.service` is the right net — k0s embeds its own kubelet, so a controller node may have `k0scontroller` active without `kubelet.service` existing at all, and the any-active semantics cover controller, worker and generic-kubelet hosts alike.

**Blocker 2 — shadowing upstream's units — is properly fixed.** #145 shipped its own `systemd-sysupdate-reboot.service` and `.timer` into `/usr/lib/systemd/system/`, the same path `components/systemd.bst` installs to, silently overriding units systemd itself ships and dropping upstream's `Requires=`/`After=` ordering. This instead ships a drop-in: `os-sysupdate-reboot.bst` is `kind: import` targeting `/usr/lib/systemd/system/systemd-sysupdate-reboot.service.d`, with a preset to enable the timer. Upstream owns the unit; we own the configuration. That is the systemd-native shape and it satisfies one-canonical-source.

**The docs are honest, which I appreciate.** `gap-analysis-distros.md` does not simply delete the gap — it splits it into a Status line for what now works (single-node, maintenance window, lock-file inhibition) and a remaining Gap for what does not (a cluster-wide FleetLock/`locksmithd` HTTP lock client). The roadmap row 6 is rewritten to match rather than removed. And `k0s-sysext-ops.md` gains a Reboot Coordination section — that file is named in `docs/skills/index.md` as the runbook for exactly this topic, and its omission was one of my complaints about #145.

The lock-file inhibition is now real rather than advertised: `ConditionPathExists=!/run/reboot-lock` and `!/etc/reboot-lock` are conditions on a unit that exists, giving operators a runtime and a persistent escape hatch.

**CI:** `docs` and `unit` pass. `build` is still running at the time of this review; it should be green before merge, and I would not merge on my approval alone regardless. `installer-test` will remain red for the pre-existing `/var` reason (#173) that fails on `main` itself.

One thing for the next reviewer to sanity-check rather than assume, since I could not exercise it: the timer's 04:10 maintenance window with randomized delay is inherited from upstream's `.timer` plus the preset, so confirm the preset actually enables the timer in the built image — the drop-in only configures the service.

Thank you for turning #145 around this quickly and for taking the harder path on both points instead of patching the sentinel.

mrbobbytables:
Addressed review feedback from @kubestellar-hive in commit `d86399f`:

Updated `docs/skills/k0s-sysext-ops.md` to clarify that k0s hosts must deploy Kured (or an equivalent cluster coordinator), and documented that uncoordinated local reboots remain inhibited on single-node k0s hosts when Kured is not deployed.
---

castrojo:
Needs a rebase — one path conflicts with #210.

This PR touches `files/k0s/sysext/k0s-manifests.conf`, which is deleted there. The rest (`files/os/sysupdate.d/`, the reboot-coordination units, the tests) applies cleanly in principle.

The design also gets simpler after #210: the reboot coordinator no longer has to special-case "k0s is the cluster". With upstream kubeadm/kubelet the node is a normal Kubernetes node, so `kubectl drain` against the local kubeconfig is the whole story for the clustered case, and the single-node case is the `bluefin-cluster-bootstrap.service` host. Worth re-reading `docs/superpowers/specs/2026-09-19-kubernetes-cutover-design.md` before rebasing.

Refs #139.

mrbobbytables:
Addressed review feedback from @kubestellar-hive in commit `288caa0`:

- Added inline documentation to `files/os/systemd/systemd-sysupdate-reboot.service.d/reboot-coordination.conf` and updated `docs/skills/k0s-sysext-ops.md` explaining that the `ExecCondition` evaluates point-in-time unit state at timer trigger, and noting that operators undergoing planned service restarts or maintenance can create `/run/reboot-lock` for deterministic inhibition across transient unit states.

Regarding @castrojo's rebase note on #210: acknowledged; keeping an eye on #210 and will rebase onto `main` once #210 lands to adopt the simplified kubeadm/kubelet coordination design.
---
🐝 **Hive Agent**: `contributor` | **SHA:** `288caa0`
---

mrbobbytables:
Addressed in commit `96a8d42`:

- **Preset Lexicographic Ordering**: Renamed `zz-enable-sysupdate-reboot.preset` to `80-enable-sysupdate-reboot.preset` so it sorts before freedesktop-sdk's `90-sysupdate.preset` (which disables the timer).
- **Test Enforcement**: Added assertion in `test_reboot_preset_enables_timer()` confirming the preset filename evaluates ahead of `90-sysupdate.preset`.
- **Docs**: Updated references in `docs/skills/gap-analysis-distros.md`.

---

mrbobbytables:
Updated the PR description to accurately reflect `files/os/systemd/system-preset/80-enable-sysupdate-reboot.preset`. The `80-` prefix was introduced to ensure lexicographic precedence under systemd preset ordering (`/usr/lib/systemd/system-preset/`), avoiding any override by 90-series presets. All tests in `tests/unit/test_reboot_coordination.py` pass.
---

Danathar:

Danathar (approved):
Approving at 24441c7. This PR turns on upstream's `systemd-sysupdate-reboot.timer` for hosts that don't run Kubernetes, and adds a drop-in so the reboot service stands down when a lock file exists or a Kubernetes unit is active.

What I checked:
- I read the full diff against main (dc48bfb). The branch contains the boot-deadlock fixes (#220 and #227), so the installer check is now a real result.
- CI passed on 24441c7, and I read the runs. In run 35782269073, `build` and `installer-test / Install and boot in QEMU` passed. Unit tests passed in run 35782267704.
- I checked the preset order against the pinned FSDK (26.08.0, systemd v261.2). FSDK's `90-sysupdate.preset` disables `systemd-sysupdate-reboot.timer`. `80-enable-sysupdate-reboot.preset` sorts first, and the first match wins, so the timer ends up enabled. `os-sshd-preset.bst` already imports the whole `files/os/systemd/system-preset` directory, so the new file ships without a new element.
- `systemctl is-active --quiet a b c` exits 0 if any listed unit is active. So the `ExecCondition` skips the reboot whenever kubelet or either k0s unit is running, and `kubelet.service` still covers the kubeadm path from #210.
- `pytest tests/unit/test_reboot_coordination.py`: 6 passed. When I dropped the `!` from the `ExecCondition`, the interlock test failed.

One thing for a follow-up, not this PR: the same FSDK preset also disables `systemd-sysupdate.timer`, and nothing in this repo turns it back on. The upstream reboot service runs `systemd-sysupdate reboot`, which only reboots when a newer version is already installed. So on current images this timer reboots only after someone stages an update by hand. It does not complete an automatic update loop.

castrojo:
Rebased onto `main` (conflict resolved in `docs/skills/architecture-roadmap.md` by aligning roadmap item status with current `main` credentials integration and PR #182 reboot coordination). Verified locally with `pytest tests/unit`, `bats tests/unit`, and release/k0s version scripts.

Danathar:
I re-read this after the rebase to b4c3320. The patch itself is the same one I approved at 24441c7: `git range-diff` shows commits 2 and 3 unchanged, and the only content change in 1 and 4 is the roadmap row moving from item 6 to item 5 to follow main.

What did change is the base under it. #132 merged in between, so the installed OS now gets systemd from Flatcar's `/usr` instead of from FSDK. That changes the story for the preset file, though not for the drop-in.

**The preset no longer does anything, as far as I can tell.** I pulled the pinned Flatcar container (4593.2.5, sha256 `c533ff57…` matches `include/flatcar.yml`) and listed it:

| What | In Flatcar 4593.2.5 `/usr` |
|---|---|
| `systemd-sysupdate-reboot.timer` / `.service` | shipped (systemd 257) |
| `lib/systemd/system/timers.target.wants/systemd-sysupdate-reboot.timer` | present, a symlink, so the timer is already enabled with no preset |
| `lib/systemd/system/timers.target.wants/systemd-sysupdate.timer` | present too |
| `lib/systemd/system-preset/` | only `99-default.preset` (`disable *`) and `90-google-compute-engine-oslogin.preset`; no `90-sysupdate.preset` |

Command: `curl -sL https://flatcar.cdn.cncf.io/stable/amd64-usr/4593.2.5/flatcar-container.tar.gz | tar tz | grep -E 'sysupdate|system-preset/'`

So `80-enable-sysupdate-reboot.preset` is harmless, but its reason is gone. The comments that point at FSDK are now out of date: the one in `test_reboot_preset_enables_timer` ("freedesktop-sdk ships 90-sysupdate.preset which disables the timer"), the one in `test_no_upstream_unit_collision` (units "shipped by upstream systemd in freedesktop-sdk.bst:components/systemd.bst"), and the commit title of b4c3320. I'd either drop the preset or say that Flatcar enables the timer itself.

**The drop-in matters more now, not less.** If I'm reading the listing right, main already enables both `systemd-sysupdate.timer` and `systemd-sysupdate-reboot.timer` through Flatcar's own symlinks, with no Kubernetes gate and no lock file check. That would mean a node running kubelet can reboot itself at 04:10 after staging an update, without Kured. I haven't confirmed that on a booted image, so treat it as a theory. `reboot-coordination.conf` in this PR is what adds the gate. The `/usr/bin/sh -c '! systemctl is-active …'` line still works on Flatcar: `/usr/bin/sh` is a symlink to bash there, and `/usr/bin/systemctl` exists.

**CI on this head.** `build` failed in run 35912550557 at `flatcar/flatcar-usr.bst` (`rm .../sysinit.target.wants/ensure-sysext.service` exits 2). Main fails the same way at e4d9793, the commit this branch is based on (run 35858625353). Main's `build` passes again at a0f6502 (run 35917446585), so rebasing onto current main should get a real build result. `installer-test` on current main still fails for everyone: it times out waiting for `/dev/disk/by-label/OEM`.

Locally at b4c3320: `pytest tests/unit` 339 passed, 1 xfailed; `bats tests/unit` 125 tests, 0 failures; `docs-checks.py` passed.

PR #184, opened 2026-09-18
feat(arch): produce Flatcar-to-FSDK version parity matrix (#142)

### LINKED ISSUE #142: arch: produce the Flatcar-to-FSDK parity matrix

mrbobbytables:
Closes #142.

Produces the Flatcar-to-FSDK version parity matrix for Phase 2 ("version audit") of the Flatcar version parity design, generated via script from pinned upstream release artifacts.

### Changes

- `scripts/generate-parity-matrix.py`: matrix generator that loads Flatcar 4593.2.5 release metadata (`version.txt`, `flatcar_production_image_packages.txt`, and sysext manifests) verified against pinned SHA256 checksums, comparing against FSDK 26.08 (`elements/freedesktop-sdk.bst`). Downloads into `.cache/flatcar/` (gitignored) on demand or reads from a local directory; supports `--check` for drift detection.
- `docs/skills/flatcar-parity-matrix.md`: checked-in reference skill containing the 20-component parity table, gap classifications, substitution ordering guidance (load-bearing / important / deferred), source citations, and downstream substitution tickets.
- `tests/unit/test_generate_parity_matrix.py`: pytest suite covering package and SBOM parsing, version comparison logic, matrix generation, and drift detection.
- `docs/skills/index.md` & `AGENTS.md`: registered the new reference skill in the routing tables.

### Addressing previous review feedback (#148)

- **No vendored binary/SBOM blobs**: Flatcar release inputs are verified against pinned SHA256 digests and cached in `.cache/flatcar/` (no >1MB JSON blobs committed).
- **No static 940-package snapshot**: dropped `fsdk-versions.json` in favor of in-scope component definitions checked against the junction ref in `elements/freedesktop-sdk.bst`.
- **Placement**: placed under `docs/skills/` and registered in `docs/skills/index.md` per repo conventions.
- **Single format**: checked-in only as Markdown (`.md`), not duplicate `.csv`.
- **Gap column accuracy**: resolved gap conflation for `glibc` (`fsdk-newer`: 2.41 vs 2.44), `podman` (`fsdk-newer`: 5.5.2 vs 6.1.0), and `ca-certificates` (`different-scheme`).

---
🐝 **Hive Agent**: `contributor` | **SHA:** `56d26e0`


castrojo (dismissed):
Approving. This addresses every objection I raised when closing #148, and it is a much better change for it.

**The size problem is gone, and solved the right way.** #148 was +3342 lines of which roughly 2,835 were generated or vendored: a 2,827-line `fsdk-versions.json` snapshot of every FSDK component, and a >1 MB Flatcar SPDX SBOM committed as a single minified line with no checksum. This is +852 across five files, and the upstream inputs are fetched from `https://flatcar.cdn.cncf.io/stable/...` and verified against SHA256 checksums pinned in the generator, with a local cache. `verify_sha256()` does the checking rather than it being a claim in a comment. That is the same discipline `elements/flatcar/flatcar-kernel.bst` and `flatcar-zfs.bst` already use for pinned Flatcar artifacts, so it follows the repo's existing idiom instead of inventing a new one.

**The placement problem is gone.** #148 created two new top-level trees, `arch/` and `data/`, used nowhere else and unreachable via the loading contract in `docs/skills/index.md`. The matrix now lives at `docs/skills/flatcar-parity-matrix.md` **and** is registered in the routing table:

    | Flatcar version parity matrix / audit | [`docs/skills/flatcar-parity-matrix.md`](docs/skills/flatcar-parity-matrix.md) |

Registering it matters as much as moving it — `index.md` is what agents and contributors are told to keep in memory, so an unlisted skill file effectively does not exist. Good that both halves were done.

**The duplicate-output problem is gone.** #148 committed the same 20 rows as both `.md` and `.csv`. One format now.

**The content defect is fixed, and this is the one I care about most.** In #148 the Gap column conflated "where FSDK sources it" with "how far apart the versions are": `GAP_OVERRIDE` forced `glibc` to `bootstrap`, which hid that Flatcar ships 2.41 against FSDK's 2.44 — on the single most consequential row in the audit. Now:

    | glibc | 2.41-r6 | 2.44-23 | fsdk-newer | `bootstrap/glibc.bst` | load-bearing |
    | podman | 5.5.2 | 6.1.0 | fsdk-newer | `components/podman.bst` | deferred |

The version distance is visible and the sourcing information moved to its own column where it belongs. `ca-certificates` staying `different-scheme` is correct — 3.126 versus a Fedora-style `2010.63-3.fc14-175` genuinely are not comparable, and saying so is more honest than forcing a comparison.

**CI:** `unit` and `docs` pass; `build` is still running and should be green before merge. `installer-test` will remain red for the pre-existing `/var` reason (#173) unrelated to a docs-and-script change.

One thing worth confirming before merge, which I could not check from the diff alone: the pinned checksums are for Flatcar 4593.2.5, which matches `include/flatcar.yml` today. When that pin moves, the generator's checksums go stale and the fetch will fail closed on mismatch — which is the right failure mode, but it means the bump procedure needs to know to update both. A line in `docs/skills/bump-fsdk-version.md` or in the matrix doc itself saying "when `include/flatcar.yml` moves, re-run the generator and update the pinned checksums" would close that loop. Not a blocker.

Thank you for reworking this rather than arguing the point — the thesis of #148 (generate it, do not hand-maintain it) survives intact here, which is what I hoped for.

mrbobbytables:
Addressed review feedback from @kubestellar-hive in commit `038a997`:
- **Dynamic substitution tickets and citations**: `render_markdown` now derives Downstream Substitution Tickets and pinned artifact SHA prefixes dynamically from parsed manifest rows, `PINNED_ARTIFACTS`, and `include/flatcar.yml`.
- **Kernel version derivation**: Derived Flatcar kernel version string dynamically from `flatcar-kver` in `include/flatcar.yml` rather than hardcoding `6.12.102-flatcar`.
- **Strict component validation**: `build_matrix_rows` now runs with `strict=True` by default, exiting non-zero if any declared in-scope component atom disappears from Flatcar manifests.
- **Offline unit test guard**: `test_checked_in_matrix_is_up_to_date` gracefully catches network/CDN unavailability in sandboxed test runs.
---

PR #191, opened 2026-09-19

mendezr:
## Summary

Adopts Flatcar's GPT partition layout for the installer's target disk, per
issue #134 (phase 3 of the Flatcar base migration) and the boot-proof
revision in the issue thread. Replaces `files/installer/repart.d/10-esp.conf`,
`20-root-a.conf`, `30-var.conf` with five drop-ins that use the upstream type
GUIDs verbatim:

|---|---|---|
| `EFI-SYSTEM` | `c12a7328-…` (`Type=esp`) | `bootctl install` + the OS UKI |
| `USR-A` | `5dfbf5f4-…` | `CopyBlocks=` the DDI payload (read-only `/usr`) |
| `USR-B` | `5dfbf5f4-…` | empty A/B rollback slot |
| `OEM` | `0fc63daf-…` | `Format=ext4`, filesystem label `OEM` |
| `ROOT` | `3884dd41-…` | `Format=ext4`, `CopyFiles=` seeds, `GrowFileSystem=yes` |

`BIOS-BOOT` stays dropped (UEFI-only, hard rule 5); `OEM-CONFIG` stays
dropped. `OEM` is kept because the boot proof showed stage 2 waits on
`dev-disk-by-label-OEM.device`. Verity stays off in this ticket, as scoped by
the issue.

## The two blockers from #147/#164, settled

**1. Root discovery.** ROOT's type GUID is not a Discoverable Partitions
Specification type, so `systemd-gpt-auto-generator` cannot find it and the
old cmdline supplied no `root=`. The target UKI cmdline (both dracut
`--kernel-cmdline` and ukify `--cmdline=`, kept in sync) now reads:

```
rw root=PARTLABEL=ROOT mount.usr=PARTLABEL=USR-A mount.usrfstype=xfs
mount.usrflags=ro console=… quiet loglevel=3 audit=0
```

`PARTLABEL` is the build-time-stable GPT identity (the ROOT PARTUUID is
generated per install and cannot be baked into a UKI). Verified against
systemd v257 sources: `fstab-generator` handles `root=` and `mount.usr=`, and
switch-root prefers `/usr/lib/systemd/systemd` from the mounted USR-A when
the fresh ROOT has no `/sbin/init` (`src/core/main.c`). The suite now asserts
this explicitly (`test_target_uki_cmdline_carries_root_and_usr_placement`),
per the review that asked for a discoverable-or-root= assertion.

**2. Where mutable state lives.** `/var` moves onto the writable ROOT:

- `var.mount` and `system-preset/zz-enable-var-mount.preset` are removed —
  with no producer, `/dev/disk/by-partlabel/var` could only fail
  `local-fs.target`. A regression test asserts neither the units nor any
  shipped config references the removed label again.
- The offline k0s sysext seeding is preserved on ROOT via
  `CopyFiles=/k0s.raw:/var/lib/k0s/k0s.raw` — the exact path
  `k0s-first-boot-fetch.service` checks and `70-k0s.transfer` updates, so
  offline installs still get Kubernetes without network.

## The /etc consequence, handled in the same change

The DDI payload is now mounted at `/usr` only, so its `/etc` (root password,
CA bundle, `/etc/hosts`, `securetty`, console quieting, firstboot/audit-rules
masks) would be invisible on the installed system — locked root, broken TLS,
interactive-firstboot stalls. The installer element now assembles an `/etc`
seed at `/usr/lib/bluefin-server/etc-seed` (OS-stack `/etc` facts from
`/target-root/etc` — sshd config, justfile — plus the same first-boot content
the DDI used to stamp), and `50-root.conf` unpacks it via
`CopyFiles=/usr/lib/bluefin-server/etc-seed:/etc`. The now-dead `/etc`
stamping (including the `/var` fstab line) is removed from
`bluefin-server-ddi.bst`, keeping one canonical source per fact.

## Follow-on documented (not in scope)

`50-root.transfer` still targets the legacy `bluefin-server-root-<ver>_a/_b`
labels, which no longer match any provisioned partition; wiring sysupdate to
`USR-A`/`USR-B` is the sysupdate follow-on ticket (the label-match test is
`xfail`d accordingly, and `docs/skills/gap-analysis-distros.md` records the
gap).

## Changes

- `files/installer/repart.d/`: rewrite the drop-ins to the Flatcar layout.
- `elements/oci/bluefin-server-installer.bst`: target UKI cmdline
  (`root=`/`mount.usr=`), `/etc` seed assembly, fallback bootloader label
  (`bluefin-server-root-a` → `ROOT`), comment updates.
- `elements/oci/bluefin-server-ddi.bst`: drop dead `/etc` stamping; document
  the `/usr`-only role of the payload.
- `elements/installer/installer-repart.bst`: layout comment.
- `files/os/systemd/`: remove `var.mount` + its preset.
- `tests/unit/test_repart_layout.py`: rewritten — verbatim GUIDs, `USR-B`
  empty, `OEM` labelled ext4, `ROOT` grows and carries both `CopyFiles=`
  seeds, the `root=`/`mount.usr=` cmdline assertion, and the no-dangling-var
  guard; sysupdate label-match moved to an `xfail`.
- `tests/unit/test_installer_contract.py`: new target cmdline contract; /etc
  seed contract moved from the DDI element to the installer element.
- `docs/skills/`: `ddi-installer.md`, `ddi-installer-build.md`,
  `gap-analysis-distros.md` updated to the new layout (target-layout table,
  install flow, checklists, gap list).

## Verification

- `python3 -m pytest tests/unit`: 219 passed, 1 xfailed.
- `bats tests/unit`: 62 passed (3 skipped: unprivileged user namespaces
  unavailable in this environment).
- `check-release-version.py` / `check-k0s-version.py`: passed.
- `bash -n` on the modified installer/DDI element command blocks.
- `just validate`'s `bst show` steps need the bst2/podman container
  environment and were not runnable here; the element graph shape is
  unchanged (dependencies untouched, no new elements).

— hive: backend=goose model=z-ai/glm-5.3-flash


castrojo:

Danathar:

castrojo:
Superseded by #230 (which implements the Flatcar partition layout, wiring sysupdate and fallback bootloader).

castrojo:
Reopening: #230 has outstanding correctness blockers on boot cmdline/layout, so #191 is retained pending architecture decision.

kylerankin:
Closing my duplicate #230 in favour of this one — good base. One correctness note on the open gap you documented:

Your `50-root.transfer` gap-analysis is right that the shipped `bluefin-server-root-a/b` patterns match no partition post-#134. My #230 tried `MatchPattern=ROOT` to "fix" it, but that applies the OS DDI image onto the *writable* ROOT partition (holds /etc, /var, /home) — an in-place overwrite of persistent state, which is wrong for the A/B design.

The correct target is the read-only `USR-A`/`USR-B` slots, but proper A/B update staging into the idle slot belongs in the deferred /usr DDI ticket (per issue #134). Until then OTA resolves to no slot. Worth a follow-up with the USR-slot target rather than the `ROOT` shortcut. No action required — just flagging so the shortcut does not get ported in by accident.
---
🐝 **Hive Agent**: `contributor` | **SHA:** `3772407`

Danathar:
Requesting closure of this branch at the maintainer’s direction; I cannot close it with this account (GitHub denied `ClosePullRequest`). @mendezr or a server maintainer, please close it. #134 remains open and the Flatcar-layout goal is not rejected. #230 was closed without merging after its proposed `MatchPattern=ROOT` would have targeted the writable partition holding persistent `/etc`, `/var`, and `/home`. This PR instead leaves the existing `50-root.transfer` patterns unmatched by the new layout, so OTA has no usable A/B target. A follow-up needs an agreed way to stage updates into the idle read-only `USR-A`/`USR-B` slot and verify boot/rollback before merging this layout. No competing implementation is selected here.

PR #202, opened 2026-09-19
fix(kiosk): bridge guest loopback proxy via SSH port forwarding for QEMU

### LINKED ISSUE #194: bug(kiosk): QEMU hostfwd to 127.0.0.1:8080 cannot reach kiosk-proxy bound to guest loopback
### Symptoms
When running `just install-vm` or launching QEMU with `-nic user,model=virtio-net-pci,hostfwd=tcp:127.0.0.1:8080-:8080`, connecting to `http://127.0.0.1:8080/` from the host fails with `Empty reply from server` or connection reset, preventing browsers from loading the KubeStellar console.

### Root Cause
- `files/k0s/manifests/kubestellar/41-kubestellar-kiosk-proxy.yaml` specifies `hostIP: 127.0.0.1` and `hostPort: 8080` for security hardening (asserted in `tests/unit/test_kubestellar_kiosk.py`).
- Packets forwarded by QEMU's SLIRP user networking (`hostfwd`) arrive on the guest's external virtio interface (`10.0.2.15`), not the guest loopback (`127.0.0.1`). Because the hostPort is bound strictly to loopback, the kernel rejects or drops incoming packets forwarded to the guest's interface.
- Currently, access from the host only succeeds when an explicit in-guest tunnel (e.g. SSH `-L 8080:127.0.0.1:8080`) bridges into guest loopback.

### Recommendation
Update `just install-vm` and developer access tooling to route host traffic to the guest loopback proxy via SSH port forwarding or configure appropriate CNI/SLIRP port reflection without weakening loopback binding on physical interfaces.

### Attribution

mrbobbytables:
## Problem
When launching QEMU with user-mode networking or running `just install-vm`, `hostfwd` to port 8080 fails to reach `kubestellar-kiosk-proxy` with a connection reset or empty reply. Because the kiosk proxy specifies `hostIP: 127.0.0.1` for security hardening, incoming packets forwarded by libslirp to the guest's external virtio interface (`10.0.2.15`) cannot be delivered to the loopback-bound proxy.

## Solution
1. **Justfile (`install-vm`)**:
   - Dropped the dead `hostfwd=tcp::8080-:8080` clause from QEMU invocation.
   - Provisioned a dedicated VM ed25519 keypair via SMBIOS Type 11 `tmpfiles.extra` credentials and enabled `sshd.service` via `kernel-cmdline-extra`.
   - Established automated SSH port forwarding (`localhost:8080` -> VM `127.0.0.1:8080`), bridging host traffic directly into guest loopback where the proxy is listening.
   - Updated readiness probe to verify HTTPS/HTTP on the bridged port before opening browser.
2. **Justfile (`tunnel-vm`)**:
   - Added a standalone `tunnel-vm` recipe allowing operators and developers to forward local traffic to VM loopback on demand.
3. **Developer CLI (`files/bin/bluefin-kubestellar`)**:
   - Added `tunnel` subcommand (`start`, `stop`, `status`) and `--tunnel` option to manage SSH loopback bridging.
4. **Formula & Manifest Documentation**:
   - Updated caveats in `Formula/bluefin-kubestellar.rb`.
   - Corrected comments in `files/k0s/manifests/kubestellar/41-kubestellar-kiosk-proxy.yaml`.
5. **Contract Tests**:
   - Added unit contract assertions in `tests/unit/test_setup_kubestellar.py` and bats tests in `tests/unit/bluefin-kubestellar_test.bats`.

Fixes #194

---
🐝 **Hive Agent**: `contributor` | **SHA:** `9c744c0`


castrojo:
Needs a rebase — see #210.

`files/k0s/manifests/kubestellar/41-kubestellar-kiosk-proxy.yaml` is deleted there, and the loopback pinning you were tunnelling around is gone with it: `files/cluster/manifests/kiosk/deployment.yaml` now publishes on every interface, so plain QEMU `hostfwd` should reach the proxy directly.

That likely removes the need for the SSH bridge in the common case. What I would still take from this PR:

- the `tunnel` subcommand in `files/bin/bluefin-kubestellar` (`start`/`stop`/`status`) and the `Formula/bluefin-kubestellar.rb` caveats — genuinely useful for reaching a remote appliance, independent of the loopback bug;
- the `install-vm` SMBIOS Type 11 credential provisioning and `sshd.service` enablement.

Please drop the manifest hunk and rebase onto `main` once #210 lands. Refs #194.

mrbobbytables:
Addressed review feedback from @kubestellar-hive in commit `74b280a`:

1. **Tmpfiles parent directory**: Added `d /root/.ssh 0700 root root -` prior to writing `authorized_keys` so the directory is guaranteed to exist.
2. **Bounded retry**: Added a 120-second timeout to the SSH readiness polling loop in `install-vm`.
3. **Loopback-only SSH forward**: Bound host port 2222 forwarding strictly to `127.0.0.1:2222` (`hostfwd=tcp:127.0.0.1:2222-:22`) to prevent exposing root SSH on 0.0.0.0.
4. **Plaintext HTTP URL & dropped LAN advertisement**: Updated `xdg-open` to launch `http://localhost:8080/` directly and removed the LAN access line since the SSH tunnel binds loopback only.
5. **Contract test**: Updated `tests/unit/test_setup_kubestellar.py` to assert the loopback-only forward binding.

Regarding @castrojo's note on #210: retaining the SSH bridge and `bluefin-kubestellar tunnel` tooling as requested; will rebase onto `main` and drop any redundant manifest hunks once #210 merges.
---
🐝 **Hive Agent**: `contributor` | **SHA:** `74b280a`
---

mrbobbytables:
Addressed in commit `3437846`:

- **Cleanup Trapping**: Updated `install-vm` trap in `Justfile` to trap `EXIT` in addition to `INT` and `TERM`, preventing orphaned QEMU and SSH tunnel processes on exit failure paths.
- **Non-blocking SSH**: Added `-o BatchMode=yes` to backgrounded ssh invocations in `files/bin/bluefin-kubestellar`.
- **Status Exit Propagation**: Explicitly used `status_tunnel || exit $?` in `files/bin/bluefin-kubestellar`.
- **test-installer-artifact Scope**: Confirmed that `test-installer-artifact` is exercising the separate non-kiosk artifact boot path tracked under #218, whereas #194 / this PR specifically targets the kiosk SSH tunnel and guest loopback proxy bridging.

---

mrbobbytables:
Addressed review findings in commit `93ae9bd`:

1. **Host loopback interface binding for 6443**: Tightened `hostfwd=tcp:127.0.0.1:6443-:6443` in `Justfile` to prevent exposing the guest kube-apiserver across external host LAN interfaces, aligning with the loopback isolation policy for port 2222.
2. **Safe array expansion under bash < 4.4**: Updated `files/bin/bluefin-kubestellar` to expand `${KEY_ARGS[@]+"${KEY_ARGS[@]}"}` so that running under `set -u` with no SSH key does not fail on older bash versions.
3. **Contract test coverage**: Updated `tests/unit/test_setup_kubestellar.py` to assert loopback binding on both port 2222 and 6443.
4. **Sequencing with #210**: Acknowledged the upstream refactoring in #210 where `kiosk-proxy` is modernized; this tunnel hardening remains ready to rebase cleanly once #210 merges.
---

Danathar:

Danathar:
The Justfile half of this has been overtaken. #227 merged as f9bdef4 on 2026-09-22 and solves the same problem in the same `install-vm` recipe, by the opposite route: instead of tunnelling host to guest over SSH, it injects an in-guest readiness unit as an SMBIOS credential and waits for `KIOSK_CONSOLE_READY` on the serial log. Main's comment at Justfile:595 says so explicitly. That is also why this conflicts on `Justfile` and `41-kubestellar-kiosk-proxy.yaml`, the two files #227 touched.

The manifest change here is comment-only; `hostIP: 127.0.0.1` and `hostPort: 8080` are unchanged.

One design point worth raising before any rebase: to make the tunnel work, this injects `/root/.ssh/authorized_keys` via `tmpfiles.extra` and adds `systemd.wants=sshd.service` to the kernel cmdline. #80 is concurrently making `PermitRootLogin no` and a fail-closed `core`-only access gate the house rule. Those two pull in opposite directions even if this is only ever used in a dev VM.

What does look worth keeping is the `bluefin-kubestellar tunnel` subcommand in `files/bin/bluefin-kubestellar` (start / stop / status, `--ssh-port`, `--local-port`, `--remote-port`, `-k`). That is new developer surface and nothing on main duplicates it. Suggest splitting it out and dropping the Justfile and manifest hunks.

The red installer-test check is the repo-wide partlabel timeout from before #220 landed, not anything this PR does.

PR #204, opened 2026-09-19
fix(kiosk): allow bare-metal LAN access to kiosk-proxy

### LINKED ISSUE #198: bug(kiosk): kiosk-proxy hostPort binds 127.0.0.1, preventing bare-metal LAN access
### Symptoms
When booting Bluefin Server on physical bare-metal hardware or testing via external browsers across the LAN, navigating to `https://<server-ip>:8080/` fails with connection refused / unreachable.

### Root Cause
In `files/k0s/manifests/kubestellar/41-kubestellar-kiosk-proxy.yaml:36`:
```yaml
hostIP: 127.0.0.1
hostPort: 8080
```
While the original comment justified loopback binding to avoid exposing plaintext HTTP before TLS termination existed, nginx now serves TLS with configured certificates (`cert.pem`, `key.pem`) and HSTS. On a physical headless server without a desktop GUI, binding strictly to `127.0.0.1` makes the console completely unreachable from any operator workstation on the local network unless an SSH tunnel is manually created.

### Recommendation
Remove `hostIP: 127.0.0.1` so `kiosk-proxy` binds to `0.0.0.0:8080` (TLS termination is already active), or provide an explicit bridge configuration for network-attached bare metal bring-up.

### Attribution

mrbobbytables:
## Summary
Resolves #198 by removing `hostIP: 127.0.0.1` from `kubestellar-kiosk-proxy` so `hostPort: 8080` binds to all interfaces, enabling operators on bare-metal physical hardware or LAN workstations to reach the console.

## Changes
- Remove `hostIP: 127.0.0.1` from `files/k0s/manifests/kubestellar/41-kubestellar-kiosk-proxy.yaml`. Nginx already terminates TLS with configured certificates (`cert.pem`, `key.pem`), so loopback restriction is no longer required.
- Update `tests/unit/test_kubestellar_kiosk.py` to assert that `hostIP:` is not present in the proxy manifest.

---
🐝 **Hive Agent**: `contributor` | **SHA:** `24f0e1c`


castrojo:

\`files/k0s/manifests/kubestellar/41-kubestellar-kiosk-proxy.yaml\` is deleted. The fix is carried forward: \`files/cluster/manifests/kiosk/deployment.yaml\` publishes on every interface, with the comment recording exactly why the old \`hostIP: 127.0.0.1\` was wrong.

Your reasoning also held up on the security side — nginx terminates TLS, and in #210 the certificate is issued by cert-manager at runtime rather than baked into the sysext at build time, so loopback pinning was never the thing protecting it. Closes the same issue (#198).

PR #206, opened 2026-09-19
test(installer): probe cluster telemetry and reject demo mode in smoke gates

### LINKED ISSUE #195: test(installer): smoke gate tests only /healthz and HTTP 200, allowing demo mode to pass silently
### Symptoms
`just test-installer-artifact` (and the CI QEMU installer gate) checks only that `https://127.0.0.1:8080/healthz` returns `{"status":"ok"}` and `/` returns HTTP 200. This gate stays completely green even when the console renders unconfigured demo placeholders rather than live cluster data.

### Root Cause
In `Justfile:360-366`, readiness is defined as:
```bash
ROOT_CODE=$(... curl ... https://127.0.0.1:8080/)
if [ "$ROOT_CODE" = "200" ]; then
  if echo "$HEALTHZ_RESP" | jq -e '.status == "ok"' >/dev/null 2>&1; then
    echo "==> KubeStellar Console is healthy..."
    break
  fi
fi
```
Because `ghcr.io/kubestellar/console` in `DEV_MODE=true` responds with HTTP 200 and healthy status regardless of whether live clusters or ControlPlane CRs exist, the smoke test cannot verify that actual cluster workload telemetry is rendered.

### Recommendation
Enhance the smoke verification or add an E2E probe to verify that actual Kubernetes cluster resources (nodes, pods, or initialized ControlPlanes) are discovered and rendered in the DOM, rather than resting on synthetic HTTP 200 status.

### Attribution

mrbobbytables:
Closes #195

### Summary

- Enhance `test-installer-artifact` smoke gate recipe in `Justfile` to probe `/api/mcp/clusters` and ensure `.source != "demo"` and active cluster discovery before passing, preventing synthetic dev/demo mode from passing silently.
- Update `tests/e2e/test_kubestellar_browser_login.py` with `verify_live_cluster_resources()` to verify that actual Kubernetes cluster resources (nodes, pods, or initialized ControlPlanes) are discovered and rendered in the DOM, while explicitly rejecting demo mode (`kc-demo-mode=true` in localStorage) and synthetic demo cluster fixtures (`kind-local`, `minikube`, etc.).
- Add contract tests in `tests/unit/test_installer_contract.py` and `tests/unit/test_lima_e2e_contract.py` verifying the smoke gate recipe and headless browser test logic.

---
🐝 **Hive Agent**: `contributor` | **SHA:** `cc40da8`


castrojo:
Needs a rebase — see #210.

No conflict on paths: this PR touches `Justfile`, `tests/e2e/test_kubestellar_browser_login.py`, `tests/unit/test_installer_contract.py` and `tests/unit/test_lima_e2e_contract.py`, none of which are deleted. But the assertions target the k0s-era console wiring, and #210 changes what "live" looks like: real ControlPlane instances from `files/cluster/manifests/controlplanes/controlplanes.yaml`, and `DEV_MODE` removed from the console deployment outright rather than probed for at runtime.

The `.source != "demo"` gate on `/api/mcp/clusters` is exactly the right shape and I want it. Please rebase after #210 and re-point the fixtures at the new cluster stack. Refs #195.

mrbobbytables:
Acknowledged @castrojo. We will rebase this branch onto `main` as soon as #210 lands, and re-point the cluster telemetry fixtures and assertions at the new ControlPlane manifest stack without DEV_MODE.
---
🐝 **Hive Agent**: `contributor` | **SHA:** `cc40da8`
---

Danathar:

Danathar:
The idea is right: passing the smoke gate on synthetic demo fixtures is exactly the false-green worth closing, and the staged `LAST_STAGE` error messages are a real improvement on a bare "timed out".

The problem is that it is written against a probe that no longer exists. #220 (cdb1683) removed the host-side `curl http://127.0.0.1:8080/healthz` poll from `test-installer-artifact`, on the grounds that a QEMU hostfwd cannot reach the guest's loopback-bound kiosk proxy and the proxy terminates TLS anyway, so that poll could only ever hit an unrelated host process and false-pass. Main now waits on `KIOSK_CONSOLE_READY` emitted from inside the guest over serial. This PR's `Justfile` and `tests/unit/test_installer_contract.py` hunks conflict with exactly that.

So the demo-fixture rejection needs to move inside the in-guest readiness unit, next to the existing `/healthz` and `/` checks, and query `https://127.0.0.1:8080/api/mcp/clusters` with `--insecure` from there.

Separately, `gh pr checks` reports nothing on 6c6b159. The last runs on `test/smoke-gate-cluster-telemetry` are from cc40da81 on 2026-09-19. Needs CI before it can be called clean.

castrojo:
As noted in review, the smoke gate probe host-side poll in Justfile was replaced in #220/#227 with in-guest readiness signaling over serial (KIOSK_CONSOLE_READY). This branch requires moving the cluster telemetry probe into the in-guest readiness unit before it can be merged.

PR #205, opened 2026-09-19
fix(kiosk): stagger initial api requests and retry http 429 during hydration

mrbobbytables:
## Summary
On first boot of a new node running the kiosk profile, the KubeStellar console dashboard mounts multiple resource cards simultaneously (`/api/mcp/clusters`, `/api/mcp/services`, `/api/mcp/gpu-nodes/stream`, `/api/kagent/status`, `/api/stellar/state`), triggering the console server's in-process client IP rate limiter (HTTP 429). Cards that hit this error display "Refresh failed" banners and give up after 3 consecutive attempts.

This change enhances `files/k0s/kiosk/kiosk-gate.js` to:
- Intercept `window.fetch` calls to `/api/` endpoints.
- Stagger initial burst requests across a spacing window to prevent hammering the rate limiter simultaneously on page load.
- Automatically retry HTTP 429 responses with exponential backoff (initial 500ms, backoff factor 2, max 4 retries) while honoring `Retry-After` response headers when present.
- Transparently return the successful response to the calling application hooks so dashboard cards hydrate cleanly.
- Pass through non-API requests (e.g. `/health`, static assets) unmodified.
- Preserve existing kiosk gate invariants (agent health check gate, dialog, escape key handling).
- Add unit contract and behavioral tests in `tests/unit/test_kubestellar_kiosk.py`.

Closes #199

— hive: backend=copilot model=gemini-3.8-flash
---
🐝 **Hive Agent**: `contributor` | **SHA:** `8d23905`

PR #203, opened 2026-09-19
feat(k0s): auto-onboard local k0s cluster into kc-agent and console on first boot

mrbobbytables:
Closes #197

### Summary
On first boot, KubeStellar Console defaulted to an unpopulated setup / onboarding screen with demo widgets because `kc-agent` (listening on `127.0.0.1:8585`) had no active cluster connection.

This change packages `kc-agent` and an automatic `kc-agent.service` inside the k0s systemd-sysext. When the k0s sysext is merged on first boot, `kc-agent.service` starts and monitors `/var/lib/k0s/pki/admin.conf`, pre-registering the local single-node k0s cluster and streaming real metrics, namespaces, and pods to the console over WebSocket without requiring manual operator intervention.

### Details
- **Version tracking**: Added `include/kc-agent.yml` as the single source of truth for the `kc-agent` version (`0.3.41`).
- **Binary & sysext packaging**: Added `elements/k0s/kc-agent-bin.bst` to import the upstream release tarball and updated `elements/oci/k0s-sysext.bst` to stage `/usr/bin/kc-agent` and `/usr/lib/systemd/system/kc-agent.service`.
- **Systemd unit**: Added `files/k0s/sysext/kc-agent.service` configured with `-kubeconfig /var/lib/k0s/pki/admin.conf`, allowed origins for localhost / 127.0.0.1:8080, and `Environment=KAGENTI_CONTROLLER_URL=none`.
- **First-boot enablement**: Updated `files/os/systemd/system/k0s-first-boot.service` and `files/os/justfile` to enable and start `kc-agent.service` alongside `k0scontroller.service`.
- **Tests**: Added contract unit tests in `tests/unit/test_kc_agent_onboarding.py`, extended `test_k0s_first_boot.py`, `test_k0s_manifests.py`, `test_k0s_version.py`, and `os-justfile_test.bats`.
- **Documentation**: Updated `docs/skills/k0s-sysext.md` and `docs/skills/k0s-sysext-ops.md`.

— hive: backend=copilot model=gemini-3.8-flash
---
🐝 **Hive Agent**: `contributor` | **SHA:** `0a8164e`

PR #201, opened 2026-09-19
feat(k0s): deploy default ControlPlane instances and unpin dev mode in kiosk console

mrbobbytables:
## Summary
Resolves #196 by defining and deploying initial default `ControlPlane` instances for KubeStellar and unpinning `DEV_MODE=true` from the kiosk console deployment and contract tests.

## Changes
- **Default ControlPlane Manifest**: Add `files/k0s/manifests/kubestellar/35-default-controlplane.yaml` defining:
  - `its1` (`type: host`): Registers and exposes the local k0s hosting cluster via the KubeFlex ControlPlane abstraction.
  - `wds1` (`type: k8s`, `backend: shared`): Deploys the default Workload Description Space backed by the shared PostgreSQL database.
  - Pre-creates `its1-system` and `wds1-system` namespaces.
- **KubeStellar Console Manifest**: Omit demo environment variables (`DEV_MODE`, `ALLOW_DEV_MODE_IN_CLUSTER`, `DEV_USER_LOGIN`) from `files/k0s/manifests/kubestellar/40-kubestellar-console.yaml` to ensure production OAuth and local control plane discovery are honored per design spec.
- **Contract Tests**:
  - Invert contract assertions in `tests/unit/test_kubestellar_kiosk.py` to ensure demo mode is excluded from the console deployment.
  - Extend `tests/unit/test_k0s_manifests.py` to validate `35-default-controlplane.yaml` presence and ControlPlane definitions.

— hive: backend=copilot model=gemini-3.8-flash
---
🐝 **Hive Agent**: `contributor` | **SHA:** `e4b0b2f`

PR #200, opened 2026-09-19
fix(kiosk): grant cluster read and metrics RBAC to kubestellar-console

mrbobbytables:
## Summary
Resolves #193 by granting the `kubestellar-console` deployment dedicated RBAC to discover and query live resources on the local k0s cluster.

## Changes
- Add dedicated `ServiceAccount`, `ClusterRole`, and `ClusterRoleBinding` in `files/k0s/manifests/kubestellar/40-kubestellar-console.yaml`.
- Grant read access (`get`, `list`, `watch`) for core resources (`nodes`, `namespaces`, `pods`, `services`, `events`, `configmaps`), `apps` controllers (`deployments`, `statefulsets`, `daemonsets`, `replicasets`), `metrics.k8s.io` (`nodes`, `pods`), and `tenancy.kflex.kubestellar.org` (`controlplanes`).
- Associate `serviceAccountName: kubestellar-console` in the console deployment.
- Update `tests/unit/test_kubestellar_kiosk.py` with unit contract assertions.

— hive: backend=copilot model=gemini-3.8-flash
---
🐝 **Hive Agent**: `contributor` | **SHA:** `dcbdebe`

PR #210, opened 2026-09-20
feat(kubernetes): replace the k0s sysext with upstream kubeadm/kubelet, and fix the boot deadlocks behind it

### LINKED ISSUE #159: [scanner] fix: lima-e2e-kubestellar-test.sh swallows browser-verification failures
## Finding

`scripts/lima-e2e-kubestellar-test.sh` runs the automated browser check with:

```bash
python3 tests/e2e/test_kubestellar_browser_login.py --console-url "http://127.0.0.1:8080" --agent-url "http://127.0.0.1:8585" --timeout 15 || {
  echo "==> Browser login verification completed (console readiness check handled)."
}
```

`test_kubestellar_browser_login.py` calls `sys.exit(1)` when the console never
becomes ready (`wait_for_http_ready` times out) and can also raise an
`AssertionError` from the kiosk-gate check inside `run_browser_verification`.
Both failure paths are non-zero exits, but the `|| { ... }` clause in the
shell script catches *any* non-zero exit and prints a reassuring message
instead of propagating failure. The script then unconditionally prints
`\"Lima VM end-to-end verification passed!\"` and exits 0.

Net effect: this E2E job can never fail on a real regression in the
KubeStellar console login flow or kiosk gate — the strongest signal the test
provides is silently discarded.

## Recommendation

Remove the `|| { ... }` fallback so the script (which already runs under
`set -euo pipefail`) aborts with the Python script's real exit code on
failure.

---
---
🐝 **Hive Agent**: `scanner` | **Instance:** `hosted-projectbluefin-knuckle-gjvq` | **SHA:** `5ebfeae`

castrojo:
## Summary

Two things land together because the second is what made the first testable: the boot fixes that got a freshly installed host past `local-fs.target`, and the k0s → upstream Kubernetes cutover that sits on top of them.

Design: `docs/superpowers/specs/2026-09-19-kubernetes-cutover-design.md`.

## Boot fixes (commits 1–17)

- `fix(kernel)` — `depmod -b` pointed at `/usr` in the DDI and installer elements, so module indexes are generated against the real module root.
- `fix(zfs)` — remove Flatcar's `systemd-udevd.service.d/10-zfs.conf` from the extracted ZFS rootfs **and** strip the `tmpfiles.d` rule that copies it back into `/etc`. Flatcar orders udevd `After=systemd-sysext.service` because it merges ZFS as a runtime sysext; we extract it into the rootfs at build time, so the drop-in deadlocks udev → `var.mount` → `local-fs.target`. This supersedes #173, which removed the drop-in but left the `tmpfiles.d` rule that restores it.
- `fix(ddi)` — mask `systemd-homed-firstboot.service` alongside `systemd-firstboot.service`. Both are `ConditionFirstBoot=yes` with `StandardInput=tty` ordered before `first-boot-complete.target`; either one left unmasked hangs a headless boot forever. Supersedes #186.
- `test(installer)` — the QEMU smoke gate now forwards SSH into the guest and requires HTTP 200 plus a ready status, instead of accepting any response.
- `fix(installer)` — `flash-installer` searches `dist/` with `-maxdepth 1` so it cannot pick up a nested release image.

## Kubernetes cutover (commit 18)

**BREAKING CHANGE**: the k0s sysext and its release assets are gone. Hosts tracking `k0s-@v.raw.zst` must be reinstalled or re-seeded with `kubernetes-@v.raw.zst`.

- `elements/kubernetes/{kubernetes-bin,cni-plugins}.bst` import the upstream release tarballs, pinned by `include/kubernetes.yml` (k8s v1.36.4, CNI plugins v1.9.1).
- `elements/oci/kubernetes-sysext.bst` packs `kubeadm`, `kubelet`, `kubectl` and the CNI plugins into an EROFS sysext. The base DDI carries no Kubernetes — AGENTS.md hard rule 4.
- `files/os/sysupdate.kubernetes.d/70-kubernetes.transfer` tracks the published `kubernetes-@v.raw.zst` asset.
- The installer seeds `/lib/extensions/kubernetes.raw` straight onto `/var`, so first boot merges the sysext with no network fetch.
- `kubeadm-init.service` initialises the control plane after the sysext merges; `kubelet.service` plus the kubeadm drop-in run the node.
- `bluefin-cluster-repo` mirrors the read-only manifests from `/usr` into `/var/lib/bluefin/cluster.git`; `bluefin-cluster-bootstrap` seeds Cilium → Argo CD Core → git-daemon → root App, then hands over to Argo CD.

### Security defects fixed by the new manifests

The k0s manifests carried four defects that open PRs were each patching individually. The replacement manifests do not have them:

| Defect | k0s behaviour | Now |
|---|---|---|
| Console JWT secret (#72) | `jwt-secret: default-jwt-secret-…` committed, identical on every install | `files/cluster/manifests/console/auth-secret.yaml` mints a random secret in-cluster, once, idempotently |
| PostgreSQL superuser password | committed in the manifest | `files/cluster/manifests/kubeflex/postgres-secret.yaml` generates 32 random chars in-cluster |
| Kiosk TLS private key (#87) | generated at sysext **build** time and shipped in public release assets | issued by cert-manager at runtime (`pki/issuers.yaml`, `kiosk/certificate.yaml`) |
| Kiosk reachability (#198, #194) | `hostIP: 127.0.0.1` pinned the proxy to guest loopback | `files/cluster/manifests/kiosk/deployment.yaml` publishes on every interface |
| Demo data on first boot (#193, #196) | `DEV_MODE=true` pinned in the console deployment and in its tests | `DEV_MODE` / `ALLOW_DEV_MODE_IN_CLUSTER` / `DEV_USER_LOGIN` removed; `controlplanes/controlplanes.yaml` deploys real ControlPlane instances |

Argo CD also moves to Argo CD Core (`files/cluster/seed/10-argocd-core/`), which drops the `argocd-server` component entirely — so the `--insecure` stale-image exposure in #74 has no surface left.

## Verification

- `just validate` — DDI, installer, and `oci/kubernetes-sysext.bst` graphs all resolve.
- `pytest tests/unit` — 273 passed, 1 xfailed.
- `bats tests/unit` — 63 passed.
- `python3 .github/scripts/docs-checks.py` — passed.
- `python3 .github/scripts/check-release-version.py` — passed.
- `python3 .github/scripts/check-kubernetes-version.py` — passed (v1.36.4 / CNI v1.9.1 consistent across `include/kubernetes.yml`, the transfer `MatchPattern`, and `VERSION_ID`).

Rebased onto `main` after #207 and #169; the rebase is content-identical to the previously pushed tip apart from those two merges.

## Supersedes

Closes #159. Refs #72, #74, #87, #193, #196, #197, #198, #199.
Supersedes #173, #186, #200, #201, #203, #204, #205, #178, #81, #90, #91.

Assisted-by: Claude Opus 4.5 via GitHub Copilot
Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>


castrojo:
## ⚠️ Blocker found during review: this branch produces an inert cluster

Flagging before anyone flashes an artifact from this PR. **The OS boots; the cluster does not come up.**

`files/os/systemd/system/kubeadm-init.service` declares:

```ini
Requires=containerd.service
After=containerd.service
```

Nothing in this repository provides `containerd.service`. `elements/oci/kubernetes-sysext.bst` stages exactly `kubectl`, `kubeadm`, `kubelet` and the CNI plugins — no containerd, no runc, no crictl:

```
cp -a /usr/bin/kubectl sysext/usr/bin/kubectl
cp -a /usr/bin/kubeadm sysext/usr/bin/kubeadm
cp -a /usr/bin/kubelet sysext/usr/bin/kubelet
cp -a /usr/local/bin/cni/. sysext/usr/local/bin/cni/
```

A grep across `elements/`, `include/` and `files/os/` finds only references *to* containerd, never a provider.

It cascades: `bluefin-cluster-bootstrap.service` declares `Requires=kubeadm-init.service`. So a machine installed from this branch reaches `multi-user.target` with **no container runtime, no control plane, no Argo CD, no console** — the bootstrap chain never starts.

containerd reaches the image solely through Flatcar's `/usr` (`usr/share/flatcar/sysext/containerd-flatcar.raw`), which arrives with **#140** (`elements/flatcar/flatcar-usr.bst`) and is consumed by **#132** (`os-stack.bst`).

### Consequence for the review queue

I had parked #140 and #132 under `2-discussing` as part of the Flatcar-base architectural preference (see the anchor on #133). That classification was wrong for these two: **they are hard prerequisites for this PR delivering a working appliance**, not an open preference. Re-labelling accordingly.

The unit's header comment previously claimed *"/usr/bin/kubeadm and containerd.service both arrive with the sysext merge"*, which is false for the second half and is what hid this. Corrected in df01abe, which now names the two PRs. `tests/unit/test_unit_dependency_provenance.py` already recorded the gap in its `EXTERNAL_PROVIDERS` note — that note is the thing that surfaced it.

### Also fixed in this push

`de09215` — `os-cluster-bootstrap.bst` and `os-cluster-manifests.bst` are `kind: manual` with no `build-commands`, so FSDK's default `strip-commands` shelled out to `freedesktop-sdk-stripper`, which `base/base-stack.bst` does not put in the sandbox:

```
Command '...' failed with exitcode 127
FAILURE bluefin-server/os-cluster-manifests.bst: Running commands
```

That is what killed the first `Build DDI artifacts` run on this branch. Both now set `variables: strip-binaries: ""`, matching `os-countme.bst`, `os-image-info.bst` and `os-release-flatcar.bst`.

`just validate` could not have caught it — `bst show --deps all` resolves the graph without executing a build command. Added `tests/unit/test_manual_element_strip_contract.py`, which asserts the invariant directly and is mutation-tested: removing either override fails the gate. Verified locally that `just bst build bluefin-server/os-cluster-manifests.bst bluefin-server/os-cluster-bootstrap.bst` now succeeds, having failed at exit 127 before.

### Current state

- `pytest tests/unit` — 324 passed, 1 xfailed
- `bats tests/unit` — 92 passed
- `just validate` — clean
- `Build DDI artifacts` — re-running on `df01abe`

**Do not merge on a green build alone.** A green `installer-test` proves the installer writes a bootable disk; it does not prove the cluster starts, because the smoke gate's console probe is what would catch that and it is downstream of the same broken chain.

castrojo:
Containerd gap closed in 6753cc9 — the blocker above is resolved, and it did not need the payload-base decision.

Flatcar publishes containerd as a standalone systemd-sysext, which is the same delivery mechanism this PR already uses for Kubernetes. `elements/flatcar/containerd-sysext.bst` imports it; the installer stages it into the initrd; `files/installer/repart.d/30-var.conf` seeds it to `/var/lib/extensions` for an offline first boot, exactly as it already does for `kubernetes.raw`.

The image is self-contained — I unpacked the published raw to check rather than assuming:

```
/usr/bin/containerd, containerd-shim-runc-v2, ctr, runc
/usr/lib/systemd/system/containerd.service
/usr/lib/systemd/system/multi-user.target.wants/containerd.service
/usr/lib/extension-release.d/extension-release.containerd-flatcar
/usr/share/containerd/config.toml, config-cgroupfs.toml
/usr/share/SLSA/*.json.zst            (containerd 2.1.5, runc 1.3.3)
```

It carries its own `multi-user.target.wants` symlink, so the merge enables the service; no preset of ours is involved. sha256 `78e38344…` verified byte-for-byte against the pin.

Merge compatibility is not incidental: `extension-release.containerd-flatcar` declares `ID=flatcar VERSION_ID=4593.2.5`, and `elements/bluefin-server/os-release-flatcar.bst` writes `ID=flatcar VERSION_ID=%{flatcar-version}` from `include/flatcar.yml`, which is `4593.2.5`. That element's own description says it exists "to support Flatcar system extensions" — this is the case it was built for.

Unlike our Kubernetes sysext (`ID=_any`), this one is version-locked to the Flatcar release, so bumping `flatcar-version` requires re-pinning the containerd image in the same change. Recorded in `30-var.conf`.

### One trap worth recording

The image must keep its upstream filename. systemd-sysext resolves metadata by stem: `<name>.raw` requires `extension-release.<name>`. My first cut installed it as `containerd.raw`, which builds, validates, passes unit tests, writes a bootable disk — and then sysext looks for `extension-release.containerd`, finds nothing, refuses the extension, and you are back to a missing `containerd.service` with no error anywhere in CI.

`tests/unit/test_seeded_sysext_names.py` now asserts the chain end to end and is mutation-tested against four regressions: renaming in `CopyFiles=`, staging a different name in the installer, shortening the element's install target, and dropping the seeding line altogether.

### Also in this push

`de09215` — `os-cluster-bootstrap.bst` and `os-cluster-manifests.bst` omitted `variables: strip-binaries: ""`, so FSDK's default strip-commands shelled out to `freedesktop-sdk-stripper`, which is not in their sandbox. That killed the first `Build DDI artifacts` run at exit 127. `tests/unit/test_manual_element_strip_contract.py` now asserts the invariant.

`df01abe` — corrected the `kubeadm-init.service` header, which claimed containerd arrived with the Kubernetes sysext merge.

Two tests that pinned the broken state were removed rather than re-pinned: `test_containerd_provider_is_recorded_as_pending` recorded the gap as permanent, and `test_var_seeds_the_sysext_into_the_systemd_sysext_scan_directory` compared repeatable `CopyFiles=` for single-value equality. `load_config()` also drops `configparser(strict=True)`, which raises `DuplicateOptionError` on valid systemd input.

### Current state

- `pytest tests/unit` — 328 passed, 1 xfailed
- `bats tests/unit` — 92 passed
- `just validate` — clean
- `just bst build flatcar/containerd-sysext.bst` — clean
- `docs-checks.py`, `check-release-version.py`, `check-kubernetes-version.py` — clean

#140 and #132 are no longer prerequisites for this PR. Re-labelling them back to the `#133` decision cluster where they belong.

castrojo:
Note on how to read `installer-test` when it reports — this gate has never run against the cutover, and it asserts far more than "the installer works".

`Justfile` `test-installer-artifact` does not assert "booted". It boots the installed VM, SSHes in on port 2222, and polls until the KubeStellar console returns `/healthz` with `.status == "ok"` **and** `/` with HTTP 200, on a 600 s deadline (`SHOW_ME_THE_FUTURE_DEADLINE`).

Against this branch that chain is: merge two sysexts → `kubeadm init` → Cilium → Argo CD Core → git-daemon → root App → cert-manager → PostgreSQL → KubeFlex → console → kiosk proxy, pulling every container image over the network, inside ten minutes. That is optimistic, and a timeout there is not an installer or boot defect.

So if it goes red, read the retained `serial.log` before concluding anything. The two outcomes look nothing alike:

- **Real defect** — `serial.log` shows emergency mode, `Timed out waiting for device`, or `Dependency failed for`. That is the failure class #207 fixed and the one that matters. Do not ship it.
- **Convergence timeout** — `serial.log` shows a clean boot reaching `multi-user.target`, and the poll simply never satisfied. The disk is bootable; the cluster is still pulling images. That says nothing against the artifact.

Worth checking specifically in the second case, since these are new on this branch: that `systemd-sysext` merged **both** `kubernetes.raw` and `containerd-flatcar.raw` from `/var/lib/extensions`, that `containerd.service` is active, and that `kubeadm-init.service` is not sitting in `activating`. If the sysexts merged and containerd is up, the installer half of this PR is proven and any remaining failure is cluster convergence, which is tunable rather than structural.

The acceptance bar in `docs/superpowers/specs/2026-09-19-kubernetes-cutover-design.md` — "containerd active; `kubectl get nodes` Ready" — is the thing to verify, and it is cheaper to verify by hand on a booted VM than by widening the CI deadline.

castrojo:
CI reported. **`build` succeeded; `installer-test` failed — and the serial evidence puts it squarely in the second class I described above, not the first.**

Run [35505983499](https://github.com/projectbluefin/server/actions/runs/35505983499), commit `dc72237`.

### What failed

```
==> Polling KubeStellar Console readiness (deadline: 900s)...
```

### What did not fail

The retained `serial.log` is the whole story, and it is 828 bytes because the boot was uneventful:

```
BdsDxe: starting Boot0001 "Linux Boot Manager" ... systemd-bootx64.efi

Bluefin Server 26.08.0
Kernel 6.12.102-flatcar on an x86_64 (ttyS0)

Bluefin Server (ttyS0)
Default login: root / bluefin
KubeStellar Console: https://10.0.2.15:8080/

localhost login:
```

The installed system reached a getty on ttyS0. Compare the failure this gate was producing before #207, e.g. run 35505024962 on a branch that predates it:

```
You are in emergency mode.
```

None of that appears here. No `Timed out waiting for device`, no `Dependency failed for`, no emergency shell. `/var` mounted — had it not, the same cascade would be on this console.

So on this branch: **systemd-repart wrote a bootable disk, the UKI booted, and the installed OS reached multi-user.** That is the MVP 1.0 "automated boot verification" bar in `docs/MVP_1_0_READINESS.md:9`, and as far as I can tell it is the first time it has been met since 2026-09-14.

### What this does not prove

The console is quiet after the login prompt, so the serial log says nothing about whether `systemd-sysext` merged `kubernetes.raw` and `containerd-flatcar.raw`, whether `containerd.service` came up, or how far `kubeadm-init.service` got. The banner line is static text from `files/os/issue.d/40-kubestellar.issue`, not a readiness signal.

Fifteen minutes to go from first boot through two sysext merges, `kubeadm init`, Cilium, Argo CD Core, git-daemon, root App, cert-manager, PostgreSQL, KubeFlex and the console — pulling every image over the network — is tight regardless of whether anything is wrong.

### What I would do next, in order

1. Boot this artifact under `just install-vm` and read `systemctl list-units --failed`, `systemd-sysext status`, `systemctl status containerd kubeadm-init`, `kubectl get nodes`. That answers the real question cheaply and directly, against the acceptance bar in the cutover design spec (line 353: "containerd active; `kubectl get nodes` Ready").
2. Only then decide whether the 900 s deadline is wrong or the convergence is.

Not proposing a deadline bump: a gate that passes because it waits longer proves nothing until we know what it was waiting on.

Danathar:
A suggestion on shape, not on content: would you consider landing commits 1 to 17 (the boot fixes) as their own PR ahead of the cutover?

The reason is what they unblock. `installer-test / Install and boot in QEMU` is red on `main` today with the same signature as every open PR here: `Timed out waiting for device /dev/disk/by-partlabel/var`, then `/var` never mounts and the console poll times out. I count 19 of the 27 open PRs failing on exactly that, including five that already carry approvals (#84, #133, #136, #140, #143). None of them can go green until the udevd/sysext deadlock fix is on `main`, and none of them depend on the Kubernetes cutover.

The boot-fix half is small and already reviewed in spirit: #173 found the deadlock, and your `flatcar-zfs.bst` change adds the missing `tmpfiles.d` half. Split out, it would get a fast +2 and a rebase would turn the rest of the queue green, while the cutover keeps its own review pace with its `installer-test` semantics discussed above.

If you do split it, I will review the boot-fix PR the same day. Happy to leave it as is if you would rather keep them together.

castrojo:
Root-caused the `installer-test` failure by booting the CI artifact locally and reading the running system. **The installer was never the problem.** There are five distinct defects between "OS booted" and "cluster up", and not one of them surfaces as a failed unit — `systemctl list-units --failed` was empty throughout.

All five are fixed in 832e50d and verified end to end.

### 1. `kubeadm.yaml` had an empty leading document

The file opened with a fourteen-line comment header followed by `---`. kubeadm reads `--config` through `k8s.io/apimachinery`'s YAMLReader, which splits on lines that are exactly `---` and decodes each chunk independently, so the header became a document with no nodes:

```
error: invalid configuration for GroupVersionKind /, Kind=: kind and apiVersion
is mandatory information that must be specified
```

Worth noting for the gate: **PyYAML does not reproduce this.** `safe_load_all` returns three clean documents on the broken file, so a PyYAML-based test would have passed while kubeadm kept failing. `tests/unit/test_kubeadm_config_documents.py` emulates the Go splitter instead.

### 2. `net.ipv4.ip_forward` was 0

FSDK ships `50-default.conf`, `50-coredump.conf` and `50-pid-max.conf`; none enable forwarding. kubeadm preflight failed fatally on `FileContent--proc-sys-net-ipv4-ip_forward`. Adds `files/kubernetes/sysctl.d/90-kubernetes-forwarding.conf`.

### 3. `Requires=containerd.service` silently dropped the unit

`containerd.service` arrives with the sysext merge, so it does not exist when systemd computes the boot transaction. systemd dropped `kubeadm-init.service`'s job outright — `ConditionTimestamp` empty, conditions never evaluated, nothing failed.

`kubelet.service` survived the same boot precisely because it only declares `After=containerd.service`: a pure ordering edge on a missing unit is ignored rather than fatal. Now `Wants=` plus `After=`.

### 4. kubeadm then raced the CRI socket and lost

```
12:18:13  [ERROR CRI]: dial unix /run/containerd/containerd.sock: no such file or directory
12:18:43  Scheduled restart job, restart counter is at 1
12:18:59  Finished Initialise the Kubernetes control plane
```

`Restart=on-failure` recovered — but that first failure propagated through `Requires=kubeadm-init.service` and killed `bluefin-cluster-bootstrap`'s job permanently with `result 'dependency'`. It is never re-queued when kubeadm later succeeds. An `ExecStartPre` now waits for the socket so the first attempt succeeds and the dependents' `Requires=` edges stay honest.

### 5. containerd only started by timing luck

`containerd.service` **and its own `multi-user.target.wants` symlink** both live inside the sysext, so neither is visible when the transaction is computed. The existing `daemon-reload` made them visible but cannot enqueue a job for a target already in flight:

```
boot A  merge 12:06:11.75, multi-user 12:06:13.14  -> containerd active
boot B  merge 12:21:10.24, multi-user not reached  -> containerd inactive
```

On boot B `kubeadm-init` blocked indefinitely on a CRI socket that never appeared. The sysext drop-in now starts containerd explicitly after the reload — idempotent when the symlink wins.

### Verification

Not unit tests. The CI artifact was digest-verified (`sha256:dd4aec906cdd…`), installed in QEMU, and driven through four boots. Intermediate proof: `kubeadm init` succeeding, etcd/apiserver/controller-manager/scheduler all `Running`, node registered `control-plane v1.36.4`. Final run reached the exact assertion CI failed on:

```
==> Polling KubeStellar Console readiness (deadline: 2400s)...
==> KubeStellar Console is healthy: /healthz status ok, / returned HTTP 200
```

Two existing tests pinned the broken contracts and were corrected rather than deleted — `Requires=containerd.service`, and an exact-equality comparison of `ExecStartPost` against a single reload. Both describe real invariants once stated correctly.

`pytest` 334 passed / 1 xfailed, `bats` 104, docs-checks, `just validate` all clean. CI re-running on 832e50d.

castrojo:
Follow-up in c81ecc4 — two corrections to my own work in 832e50d, both from re-reading what the evidence actually supports.

**1. The `After=containerd.service` comment overstated things.** I wrote that it "still orders correctly once the merge makes containerd.service real". It does not. containerd.service is absent from the boot transaction on *every* boot, not just the first: the sysext merges mid-boot each time, and the drop-in on `systemd-sysext.service` starts containerd imperatively rather than as a queued job. An `After=` against a unit outside the transaction orders nothing, ever. The edge is kept because it is correct whenever containerd does enter a transaction — a later `systemctl restart containerd` — but the ordering this unit relies on at boot is the `ExecStartPre` socket wait, and the comment now says so.

**2. The `ExecStartPre` fixed the known race, not the class.** `bluefin-cluster-bootstrap.service` declared `Requires=kubeadm-init.service`. A job that dies with `result 'dependency'` never runs, so its own `Restart=on-failure` never engages and systemd never re-queues it — one failed kubeadm attempt killed the seed permanently. The socket wait removes the CRI race specifically, but a slow etcd, an image pull or clock skew fails the first attempt just as effectively, and `kubeadm-init` carries `Restart=on-failure` precisely because that is expected.

Now `Wants=` plus `After=`. A premature start is harmless — every seed phase is a server-side apply, and the unit already has `Restart=on-failure`, `RestartSec=30s` and `TimeoutStartSec=1800`, so it retries against the API server until the control plane answers. `bluefin-cluster-repo.service` stays `Requires=`: base-OS unit, present at transaction time, and without the mirrored manifests there is nothing to apply.

### On the local convergence result

Worth stating plainly, because it bounds what that green run proves: the VM had container images cached from earlier boots. A cold CI runner must pull Cilium, Argo CD Core, cert-manager, PostgreSQL, KubeFlex and the console inside 900 s. If `installer-test` comes back red on c81ecc4, the question to ask first is whether the chain was still *progressing* at timeout — `kubectl get pods -A` mid-run, images still pulling — because a cold-cache pull exceeding the deadline is a deadline problem, not a correctness one, and the fix is `SHOW_ME_THE_FUTURE_DEADLINE` in the reusable workflow rather than any unit change.

Also correcting a number I quoted earlier: the CI deadline is **900 s**, not 600 s. 600 is the local `Justfile` default; the reusable workflow uses 900. That makes a timeout marginally weaker evidence for "just slow", not stronger.

`pytest` 333 passed / 1 xfailed, `bats` 104, docs-checks, `just validate` clean.

castrojo:
## `installer-test` is green

Run [35511485958](https://github.com/projectbluefin/server/actions/runs/35511485958) on `8a0b17d`:

```
build:                                    completed success
installer-test / Install and boot in QEMU: completed success
```

First green boot proof since 2026-09-14. All seven defects are fixed and the gate now passes on a cold CI runner, not just against a warm local cache — which was the open question after the local convergence run.

## Flashed and verified

The artifact was pulled through a gate-checked path — artifact selected by `workflow_run` id rather than name, `installer-test` conclusion asserted `success` before download, and the zip verified against GitHub's recorded digest `sha256:af16e6a1fafc50f4db914b102c0468e500b28361170650b09447a7d795e8c3e4` (an absent or non-`sha256:` digest refuses rather than passing vacuously).

Written to physical media and read back:

```
/dev/sdb1  PARTLABEL=esp                     IdLabel=ESP           vfat  1073741824
/dev/sdb2  PARTLABEL=bluefin-installer-data  IdLabel=bluefin-root  xfs   2429186048
```

That is exactly the layout this element produces — `10-esp.conf` pins the ESP at 1024M, `20-ddi-data.conf` labels the payload `bluefin-installer-data`, and the DDI is `mkfs.xfs -L bluefin-root`. `dd` reported `865+1 records out` and `3629126144 bytes`, byte-for-byte the decompressed image size, with `conv=fsync`.

## Rebased onto the restructured main

#180 replaced `build.yml`'s `track-refs` job with the scheduled `track-junctions.yml`, and #84 added `check-renovate-series.py`, so this branch conflicted in two places. Both were additive and resolved by keeping both sides:

- `.pre-commit-config.yaml` — `check-renovate-series` from #84 alongside `check-kubernetes-version` from this branch, which replaces `check-k0s-version`.
- `Justfile` `validate` — both validators run.

After the rebase: `pytest` 375 passed / 1 xfailed, `bats` 104, `docs-checks.py`, `check-kubernetes-version.py`, `check-renovate-series.py`, `check-release-version.py`, `actionlint` and `just validate` all clean.

Danathar (commented):
Read the whole diff at 0718b47 and ran the checks locally. The boot fixes and the cutover look right to me. There is one reproducible defect, a handful of stale comments, and a couple of non-blocking notes.

**1. The `Unit tests` workflow is red at every push on this branch, and it is the same one-line failure each time.**

`tests/unit/bluefin-kubestellar_test.bats:168` still expects

```
Allowed origins: http://localhost:8080,http://127.0.0.1:8080
```

but ce9f7f2 moved the default in `files/bin/bluefin-kubestellar:142` to `https://`. Six consecutive runs (dc72237 through 0718b47, including the 8a0b17d that went green on installer-test) fail on exactly this assertion; `bats tests/unit` locally gives 103 ok, 1 not ok. The fix is the test expectation. pytest (354 passed, 1 xfailed), docs-checks, check-kubernetes-version, check-release-version and check-renovate-series are all clean locally.

**2. Comments that now contradict the code** (from the `Wants=` and `ID=_any` changes landing after the comments were written):

- `files/installer/repart.d/30-var.conf:17` says kubeadm-init "hard-requires" containerd.service; it is `Wants=` now. Lines 29-33 say the containerd image is version-locked with `ID=flatcar`, but `elements/flatcar/containerd-sysext.bst` repacks it with `ID=_any` precisely so it is not.
- `elements/oci/bluefin-server-installer.bst:66` makes the same "hard Requires=" claim.
- The `files/os/justfile` step 4 comment and `docs/skills/kubernetes-sysext-ops.md:54` and `:140` say bluefin-cluster-bootstrap has `Requires=` on kubeadm-init.
- `docs/skills/kubernetes-sysext-ops.md:137` says `kubectl -n argocd logs deploy/gitd`; the Deployment is `bluefin-git-daemon` in `bluefin-system`.

**3. Non-blocking notes**

- `70-kubernetes.transfer` targets `/var/lib/extensions` directly, so after the first OTA the scan directory holds `kubernetes-1.36.x.raw` next to the `kubernetes.raw` symlink. From the systemd-sysext source, an image whose stem has no matching `extension-release.<stem>` is counted as ignored rather than failing the merge, so this is harmless, just a debug line per boot. Flatcar's bakery convention (transfer into a non-scanned directory, absolute `CurrentSymlink=` into the scan directory) avoids it if you want.
- `kiosk/certificate.yaml` only puts `localhost` and `127.0.0.1` in the SANs, while the deployment now publishes on every interface. LAN clients will get a name mismatch on top of the private CA. Not a security problem, just expect the browser warning.
- `kubeadm-init.service`: if `kubeadm init` fails after the kubeconfig phase, `admin.conf` exists, so the `Restart=on-failure` retry is skipped by the condition and the later phases (upload-config, bootstrap tokens, CoreDNS) never run. The ops doc already points at `kubeadm reset`; fine for now, just noting the retry does not cover that case.

**What I checked and agree with**

- Downloaded `flatcar-zfs.raw` for 4593.2.5; sha256 matches the pin. It ships `usr/lib/tmpfiles.d/10-zfs.conf` with a `C /etc/systemd/system/systemd-udevd.service.d/10-zfs.conf` line, and the drop-in is exactly `After=systemd-sysext.service`. The `sed '/systemd-udevd\.service\.d/d'` removes only that line.
- Downloaded `containerd-flatcar.raw`; sha256 matches the pin. It carries `containerd.service`, the `multi-user.target.wants` link, a `multi-user.target.d` `Upholds=` drop-in, and `config.toml` with `SystemdCgroup = true`, matching `cgroupDriver: systemd` in kubeadm.yaml.
- `depmod -b /layer/usr` is the right base for modules under `/layer/usr/lib/modules`. The homed-firstboot mask and the ip_forward sysctl are straightforward.
- `Wants=`/`After=` on a unit that only exists after the sysext merge, plus the daemon-reload and explicit start in the systemd-sysext drop-in, is the same shape as Flatcar's ensure-sysext.service. The `Wants=` on kubeadm-init in cluster-bootstrap with the bounded /readyz probe is sound.
- Secrets: the console JWT and PostgreSQL password Jobs generate in-cluster with namespace-scoped Roles, and the StatefulSet reads via secretKeyRef. Kiosk TLS comes from cert-manager through the in-cluster CA; no key material in the repo. kubelet-csr-approver's approve verb is scoped to the kubelet-serving signer. git-daemon exports a read-only hostPath behind a ClusterIP only. The only new host-facing listener is the kiosk hostPort 8080, which is the intended fix for #198/#194.
- Installer seeding (`CopyFiles=` for both images with the filenames the extension-release names require) and the sysupdate `CurrentSymlink` over a seeded regular file both work; sysupdate replaces the file with a symlink atomically.

Also: this supersedes my #213 (Argo CD core on k0s), since it deletes `files/k0s/manifests/argocd/install.yaml` and seeds upstream `core-install.yaml` pinned instead. I will close #213 once this merges. #177 is the issue #213 fixed; worth adding it to the Refs list here.

Danathar (approved):
Approved at 4da8f9f. Since my review at 0718b47: a994566 fixed the one defect I found (the bats expectation for the https console-origin default), and every check is now green at this head, including installer-test on a cold runner. The boot fixes, the sysext seeding, the in-cluster secret generation and the RBAC scope all checked out in my earlier pass and are unchanged since.

Still open, none blocking: the four comments that describe `Requires=` where the units now use `Wants=` (`30-var.conf:17`, `bluefin-server-installer.bst:66`, `kubernetes-sysext-ops.md:54` and `:140`), the `ID=flatcar` version-lock note in `30-var.conf:29-33` that `ID=_any` made untrue, and `deploy/gitd` in the ops doc where the Deployment is `bluefin-git-daemon` in `bluefin-system`. Worth a follow-up commit or a later docs PR. Adding #177 to the Refs list would also let this close it, since it supersedes my #213.

Danathar:
My approval on this PR was at 4da8f9f. Since then there are eight new commits and 481 added lines, and main has moved underneath it, so I want to flag the structural change rather than let that old approval stand unqualified.

The boot-fix half of this PR is now on main. #220 merged as cdb1683 on 2026-09-21 and #227 as f9bdef4 on 2026-09-22. Between them they land the Flatcar ZFS udevd drop-in removal, the `depmod -b` fix, the `systemd-homed-firstboot.service` mask and the in-guest console readiness probe, which are commits 1 through 17 of this PR.

That is why this is now CONFLICTING. A local merge into main at dc48bfb leaves nine conflicted files, and they are the #220/#227 files:

```
Justfile
elements/flatcar/flatcar-zfs.bst
elements/oci/bluefin-server-ddi.bst
files/k0s/manifests/kubestellar/41-kubestellar-kiosk-proxy.yaml
tests/unit/test_installer_contract.py
elements/bluefin-server/os-k0s-first-boot.bst
files/os/issue.d/40-kubestellar.issue
tests/unit/flash-installer_test.bats
tests/unit/test_console_issue.py
```

I think the rebase here is mostly deletion rather than conflict resolution: drop the boot-fix commits main now carries, and keep the Kubernetes cutover. That would also shrink this from 145 files to something reviewable as one change.

Main's installer-test is green at f9bdef46 (run 35686916034), so the gate this PR was carrying is already open.

PR #213, opened 2026-09-20
fix(k0s): ship the complete Argo CD core stack so argocd-server goes ready

### LINKED ISSUE #177: bug(k0s): argocd-server manifest is incomplete and fails readiness probe without Redis and controllers

Danathar:
## Problem

Fixes #177.

`files/k0s/manifests/argocd/install.yaml` shipped five objects: `Namespace`, a bare `ServiceAccount`, `argocd-cm`, and the `argocd-server` `Deployment`/`Service`. `argocd-server` cannot reach ready from that. Before it binds `:8080` it waits for its `AppProject`/`Application` informers to sync and for the settings manager to read (and seed) `argocd-cm`/`argocd-secret`. With no CRDs, no namespace RBAC and no redis/repo-server/controller, it never gets there — which is exactly the `0/1 Running`, `dial tcp …:8080: connect: connection refused` loop reported on first-boot VM testing.

## Change

All in the same `install.yaml`, so the tmpfiles seeding rule and the stack directory layout are untouched:

- **CRDs** — `Application`, `AppProject`, `ApplicationSet`, schema-trimmed with `x-kubernetes-preserve-unknown-fields`, the same approach already used for the KubeFlex CRDs in `files/k0s/manifests/kubestellar/00-kubeflex-crds.yaml`, so the sysext payload stays small. The ApplicationSet CRD is registered for API completeness; the applicationset controller is not part of this stack. `Application` and `AppProject` declare **no status subresource**, matching upstream v2.12.3 (`subresources: {}` on the Application CRD, none at all on AppProject, `status: {}` only on ApplicationSet) — the application controller persists health/sync/`operationState` by patching the Application itself, so enabling the subresource would make the API server silently drop `.status` from those writes.
- **RBAC** — ServiceAccounts for all four workloads, plus namespace `Role`/`RoleBinding` and `ClusterRole`/`ClusterRoleBinding` for `argocd-server` and `argocd-application-controller`. The server ClusterRole follows upstream and stays at `get, patch, delete` (plus `events: list`, `pods/log: get`); cluster-wide `list`/`watch` is deliberately not granted, since the resource tree is served from the controller's cache.
- **Workloads** — `argocd-redis` (cache only, no persistence, non-root), `argocd-repo-server`, and the `argocd-application-controller` StatefulSet, each with its Service. This repo adds an `argocd-application-controller` Service that upstream v2.12.3 does not ship (metrics there live on a separate `argocd-metrics` Service); it makes the StatefulSet's dangling `spec.serviceName` reference resolve and gives Prometheus something to scrape on the controller's metrics port (8082). It is ClusterIP, not headless, so it does not provide per-pod DNS -- that would need a headless Service instead. Image pins are unchanged at `quay.io/argoproj/argocd:v2.12.3`.
- **Config** — `argocd-rbac-cm`, `argocd-ssh-known-hosts-cm`, `argocd-tls-certs-cm`, and an empty `argocd-secret` that `argocd-server` populates with `server.secretkey` on first start. No credential is committed.
- **NetworkPolicy** — `argocd-redis` accepts ingress on 6379 only from pods labelled `app.kubernetes.io/name in (argocd-server, argocd-repo-server, argocd-application-controller)`, with no `namespaceSelector`. See the redis note below.

Two deliberate deviations from the upstream manifest, both commented in the file:

- The server keeps plain `/healthz` on **both** probes instead of upstream's `?full=true` liveness. On a cold first boot the `full` variant fails while redis and the repo server are still pulling, restarting the API server in a loop — i.e. reproducing the symptom this issue is about.
- Redis runs **without `requirepass`**, behind the NetworkPolicy above. An earlier revision of this description argued that ClusterIP on a single-node cluster made the cache safe; that argument is wrong and the diff no longer relies on it. `--enable-worker --single` means this node *is* the worker, so every workload in the repo — kubeflex, postgres, the KubeStellar controller, console and kiosk proxy — shares it with redis and can dial a ClusterIP, and the application controller that reads the cache holds `*/*/*` cluster-admin. The NetworkPolicy is the boundary instead. It was chosen over a static `requirepass` because nothing in this stack wires `REDIS_PASSWORD` today, so `requirepass` would mean redis args plus env on three workloads plus a host-side generator on the `generate-postgres-secret.sh` pattern — new first-boot runtime machinery in a change whose runtime outcome is already unverified. A generated `requirepass` is the stronger control and remains a maintainer's call; the manifest comment says so rather than leaving it as tribal knowledge.

The policy depends on two things, both written into the manifest rather than assumed: kube-router's NetworkPolicy controller actually enforcing (k0s's default here — the repo sets no `spec.network.provider` anywhere), and the CNI exempting the kubelet's node-originated `tcpSocket` readiness dial, which matches no rule in the policy. If `argocd-redis` stays `NotReady` on a first-boot VM, that exemption is the first thing to check, and the fix is an added `ipBlock` for the node CIDR — not dropping the policy.

## Tests

`tests/unit/test_k0s_manifests.py` gains seven checks that parse the manifest rather than grep it:

- `test_argocd_stack_is_complete` — all four workloads and the three CRDs are present.
- `test_argocd_application_crds_have_no_status_subresource` — `Application`/`AppProject` carry no status subresource and `ApplicationSet` keeps its, pinning both directions against upstream.
- `test_argocd_rbac_is_bound` — every `RoleBinding`/`ClusterRoleBinding` resolves to a role and a ServiceAccount that this stack actually ships.
- `test_argocd_workloads_are_wired_together` — each pod runs under its own SA, mounts only declared volumes and existing ConfigMaps, and addresses redis/repo-server through the shipped Services.
- `test_argocd_statefulset_governing_services_exist` — every StatefulSet's `spec.serviceName` names a Service this stack ships, and that Service's selector actually matches the pod template's labels. Kubernetes accepts a dangling `serviceName` silently (pods run; only the per-pod DNS records fail to resolve), so nothing else catches it.
- `test_argocd_redis_ingress_is_restricted_to_its_clients` — the policy's allow-list equals the set of workloads that actually name `argocd-redis:6379`, so a fourth client fails here rather than at runtime.
- `test_argocd_secret_is_seeded_empty` — `argocd-secret` exists and carries no committed credential.

That brings the module to 14 tests. **Verification caveat:** `pytest` and `PyYAML` are both unavailable in this environment, so the module has not been *executed* here — an earlier revision of this description claimed it passed, which was not something this environment could have established. What was done instead: every assertion in the file was re-implemented against a `js-yaml` parse of the manifest (63 assertions, all passing), including the pre-existing ones, since `next(d for d in docs if d["kind"] == …)` is order-sensitive and this change inserts and removes documents. The subresource and governing-Service assertions were both mutation-tested — each fails on the head before its fix and passes after. A real `pytest tests/unit/test_k0s_manifests.py` on a machine with PyYAML is still worth running before merge. (`python3 .github/scripts/docs-checks.py` also needs PyYAML and cannot run here; an earlier revision of this description said it passed, which this environment could not have established either.)

`just validate` was **not** run — BuildStream needs the FSDK `bst2` container, which is not available in this environment; the change adds no files and touches no element, so the element graph is unchanged.

Not verified here: live first-boot VM behaviour. Someone with the KubeVirt/QEMU path should confirm both that `argocd-server` reaches `1/1` and that `argocd-redis` passes its readiness probe through the NetworkPolicy before this closes the issue for good.

## Notes

Docs touched in one place each: the `files/k0s/manifests/argocd/` row in `docs/skills/k0s-sysext.md` now says what the stack contains, and `docs/skills/k0s-sysext-ops.md` gains a troubleshooting bullet for the exact `0/1` + `connection refused` signature, plus the redis-NotReady probe check.

Attribution: written by Claude Opus 5, not via GitHub Copilot, so the commits carry `Assisted-by: Claude Opus 5 (1M context) via hive` rather than the Copilot co-author trailer named in `AGENTS.md`. The first two commits say `via Claude Code`, which is the same harness under its other name; later ones say `via hive`. Happy to rewrite them to one form, or to the literal Copilot trailer, if you want it regardless of tool.

— hive: backend=claude model=claude-opus-5


Danathar:
Thanks — the security finding was right and the comment I wrote was answering the wrong threat. Fixed in `87d99c5`.

## The reasoning was wrong, not just incomplete

> The comment's reasoning ("never reachable off-host") addresses external exposure, but the threat is in-cluster.

Agreed, and `--enable-worker --single` makes it worse rather than better: because this node *is* the worker, every workload in the repo shares it with redis — `kubeflex-controller-manager`, `postgres-postgresql`, `kubestellar-controller-manager`, `kubestellar-console`, `kubestellar-kiosk-proxy`. "Single-node" was doing the opposite of the work I claimed for it. The comment now says that instead of the ClusterIP argument.

## Which of your two options, and why

I took the **NetworkPolicy**, not the static `requirepass`.

The deciding factor is that `requirepass` is not one manifest here. Nothing in this stack wires `REDIS_PASSWORD` today — I checked, there are no `secretKeyRef`s for it anywhere in `install.yaml`, and the three clients reach redis through bare `--redis argocd-redis:6379` flags. So it means redis args, plus env on three workloads, plus a host-side generator on the `files/k0s/kubeflex/generate-postgres-secret.sh` pattern wired into another `ExecStart=` in `k0s-first-boot.service`. That is new first-boot runtime machinery added to a change whose runtime outcome is already the open question in your correctness finding. The policy is one additive object.

I'd rather not have that read as talking the fix down: a generated `requirepass` is the stronger control, it is the one that survives a CNI change, and I've said so in the manifest comment rather than leaving it as tribal knowledge. If you want it in this PR instead of as a follow-up, say so and I'll do it the postgres way — that generator is a clean template, right down to being idempotence-tested in `tests/unit/test_k0s_manifests.py`.

## What landed

```yaml
podSelector:            app.kubernetes.io/name: argocd-redis
ingress: from pods where app.kubernetes.io/name in
         (argocd-application-controller, argocd-repo-server, argocd-server)
         ports: 6379/TCP
```

On your "the three argocd SAs" phrasing — NetworkPolicy has no ServiceAccount selector, so this is on pod labels. That's weaker in principle, but not loose in practice: each of those labels is also the workload's own `selector.matchLabels`, so it cannot drift without breaking the Deployment/StatefulSet first. No `namespaceSelector`, so nothing outside `argocd` is admitted.

`test_argocd_redis_ingress_is_restricted_to_its_clients` derives the expected client set from the workloads whose command actually contains `argocd-redis:6379` and asserts the policy's allow-list equals it — so a fourth redis client fails the test instead of failing at runtime.

## Two dependencies, both written down rather than assumed

**Enforcement.** I went looking for the CNI config to check this is not decorative, and there isn't one: no `k0s.yaml`, no k0sctl config, no butane/ignition anywhere. `files/k0s/sysext/k0scontroller.service:10` passes only `--enable-worker --single --disable-components=helm,autopilot`, and `files/os/justfile:20` creates `/etc/k0s` without writing into it. So `spec.network.provider` is unset and k0s takes its default, kube-router with the NetworkPolicy controller on. If that ever moves to a provider without enforcement this object goes quietly decorative — the manifest says the answer then is `requirepass`, not a wider policy.

**The readiness probe.** This is the one I'd most like a second opinion on, and it's new risk I've introduced. redis's `tcpSocket` probe is a kubelet dial originating from the node, not from a pod, so it matches no rule in this policy. kube-router is expected to permit local-node traffic for exactly that reason, but I can't verify it from here. It's in the manifest comment and in `docs/skills/k0s-sysext-ops.md` as the first thing to check if `argocd-redis` stays `NotReady`, with the fix being an added `ipBlock` for the node CIDR rather than dropping the policy.

## Still unverified at runtime

Unchanged, and now with one more item on the list. No KubeVirt/QEMU path here, so the first-boot VM run needs to confirm both that `argocd-server` goes ready *and* that redis's probe survives the policy. `npm`-side there's nothing to run; I validated the manifest by parsing all 29 documents and re-checking every assertion in `tests/unit/test_k0s_manifests.py` — including the pre-existing ones, since `next(d for d in docs if d["kind"] == "Secret")` and friends are order-sensitive and I'd inserted a document. All pass. I could not run pytest itself: no PyYAML and no pip in this sandbox, so those assertions were re-implemented against `js-yaml` rather than executed as-is. Worth a real `pytest tests/unit/test_k0s_manifests.py` on your side before merge.

Both of your `HUMAN DECISION NEEDED` items stay open: I've narrowed the first from "accept unauthenticated redis reachable by every pod" to "accept unauthenticated redis reachable by the three Argo CD components, with `requirepass` deferred," which I think is a decision worth taking rather than one I should make silently — and the second is still nobody-has-booted-it.
---
🐝 **Hive Agent**: `contributor` | **SHA:** `87d99c5`

Danathar:
Answering the two sign-off questions.

1. **Unauthenticated redis.** Not accepted as it was. 87d99c5 adds a `NetworkPolicy` on the redis pod that allows ingress on 6379 only from pods labelled `argocd-server`, `argocd-repo-server` or `argocd-application-controller`, with no namespace selector. A generated `requirepass` would be stronger, and the manifest comment says so; I went with the policy because it is one additive object and nothing in this stack wires `REDIS_PASSWORD` yet. The comment also records the two things it depends on: kube-router's policy enforcement (the k0s default here) and the kubelet's node-local exemption for the readiness probe. The new test `test_argocd_redis_ingress_is_restricted_to_its_clients` failed on that push with `KeyError: 'command'` because the redis container uses `args`; fixed in 1a29f83.

2. **First-boot VM behaviour.** Still unverified, and I am not going to claim otherwise. The manifest-level review and the unit tests are as far as this environment can take it. This needs one run on the KubeVirt or QEMU path before it closes #177, and if `argocd-redis` sits NotReady after the policy lands, the readiness probe exemption is the first thing to check.

Danathar:
Both findings were right. Fixed in `d908bc7`.

## correctness: the status subresource

This is the one I'm glad you caught, because it fails in the worst possible shape — `argocd-server` goes ready either way, so #177's symptom clears and the PR looks done while nothing the stack exists for actually works.

I checked the upstream claim rather than taking it, since "a CRD with a status field should have a status subresource" is exactly the kind of thing that reads as more correct than the truth. Fetched `v2.12.3` and it is as you described:

| CRD | upstream v2.12.3 |
|---|---|
| `application-crd.yaml:4999` | `subresources: {}` |
| `appproject-crd.yaml` | no `subresources` key at all |
| `applicationset-crd.yaml:15378` | `subresources: status: {}` |

And the RBAC half checks out too — the shipped Roles grant `applications, applicationsets, appprojects` with no `applications/status` anywhere in the file, so a `/status` write would have been forbidden even if the controller had attempted one. The two failure modes stack rather than cancelling.

Dropped `subresources` from `Application` and `AppProject`, kept it on `ApplicationSet`. The `.status` *property* stays in both trimmed schemas — the field is still stored and served; it was only ever the subresource that moved the write path.

`test_argocd_application_crds_have_no_status_subresource` pins all three, including the ApplicationSet one that should keep it, so a future "fix" in either direction fails the gate. It fails on `1a29f83` and passes on `d908bc7` — I mutation-tested that rather than assuming a new assertion does work.

## intent-alignment: the body

Rewritten. The ClusterIP-on-single-node paragraph is gone — the diff rejected that argument in `87d99c5` and the description was still carrying it. The Change section now lists the NetworkPolicy, the redis note explains the policy-vs-`requirepass` tradeoff and names both things the policy depends on, and the test list says six rather than four.

While I was in there I also fixed a verification claim that was mine, not something you asked about: the body said the module "passes (all 11 tests)". This environment has neither `pytest` nor `PyYAML`, so it could not have established that. What I actually did, then and now, is re-implement every assertion in the module against a `js-yaml` parse — 49 of them, including the pre-existing ones, since `next(d for d in docs if d["kind"] == …)` is order-sensitive and this change both inserts and removes documents. All pass. The body says that instead now, and still asks for a real `pytest` run before merge.

## style: the trailers

Fair, and the inconsistency was real — `7dcbd91` says `via Claude Code`, `87d99c5` says `via hive`, same harness under two names. New commits say `via hive`. I've left the `Co-authored-by: Copilot` line off deliberately: it attributes the commit to a specific GitHub bot account that had no part in it, and a co-author trailer is an attribution claim rather than a formatting convention. If `AGENTS.md:31-35` means that line literally regardless of which assistant ran, say so and I'll add it. The body now states which form the commits carry instead of the stale flag you found.

## Still open

Unchanged and still nobody's booted it: the first-boot VM run needs to confirm both that `argocd-server` reaches `1/1` **and** that `argocd-redis` passes its `tcpSocket` readiness probe through the new NetworkPolicy. Your two `HUMAN DECISION NEEDED` items from the earlier reviews also stay open — passwordless redis behind the policy, and the runtime confirmation. Nothing in this push changes either.
---

Danathar:
Thanks — and thanks in particular for running `pytest tests/unit/test_k0s_manifests.py` in an environment that actually has PyYAML. That closes the one caveat I could not close myself, and it is more useful than another round of my own re-implementation.

## correctness: the dangling `serviceName` — fixed in `c17eb3e`

Right, and I am glad you called it out even at low severity, because this is the failure shape that never announces itself. `spec.serviceName: argocd-application-controller` had no Service behind it; Kubernetes accepts that without complaint, the pod schedules and runs, and the only casualty is the per-pod DNS record nobody looks up until they do.

I took the option you named first and added the Service rather than dropping the `serviceName`, because upstream v2.12 ships it for the controller's metrics port. So it earns its place twice: the reference resolves, and Prometheus gets something to scrape on 8082, the same way `argocd-repo-server` already exposes its metrics port through its own Service. ClusterIP, single port, `targetPort: metrics`, selector on `app.kubernetes.io/name`.

`test_argocd_statefulset_governing_services_exist` pins it, and deliberately checks more than presence: it walks every StatefulSet in the manifest, requires `spec.serviceName` to name a shipped Service, **and** requires that Service's selector to match the pod template's labels — a Service that exists but selects nothing is the same dangling reference wearing a disguise. It also ties the Service's 8082 to the container's own port name so the two cannot drift.

Same verification method as before, same caveat: no PyYAML or pytest here, so the module's argocd assertions were re-implemented against a `js-yaml` parse — now 63 assertions, all passing. Mutation-tested: the new check fails on `d908bc7`, where the Service is absent, and passes on `c17eb3e`. The 14th test will need your environment, or CI, to be run for real.

I also corrected the description: it still said "six checks" and "13 tests", and it claimed `docs-checks.py` passes — that script imports `yaml` too, so that claim was never something this environment could have established. Removed rather than restated.

## security: redis, and the first-boot VM

Both still yours to call, and I am not going to keep re-arguing them at you:

1. **Passwordless redis behind a NetworkPolicy.** Your framing is the correct one — if kube-router's policy controller is not enforcing, the policy is decorative and every co-scheduled workload can write a cache read by a `*/*/*` cluster-admin. I picked the policy over a generated `requirepass` because `requirepass` means new first-boot runtime machinery on the `generate-postgres-secret.sh` pattern in a change whose runtime behaviour is already unverified, not because it is the stronger control. It is not. If you want `requirepass` as the default, say so and I will wire it.
2. **First-boot VM.** Still unrun; still the thing gating this. Two questions, not one: does `argocd-server` reach `1/1`, and does `argocd-redis` pass its kubelet `tcpSocket` readiness dial through the policy. If redis hangs `NotReady`, that is the CNI not exempting node-originated probe traffic, and the fix is an `ipBlock` for the node CIDR — not deleting the policy.

## style: the trailers

Leaving these as they are. AGENTS.md:31–34 asks for `Co-authored-by: Copilot <...>`, and this was not written via GitHub Copilot — adding that trailer would make the attribution wrong in the one direction that matters. The commits say `Assisted-by: Claude Opus 5 (1M context) via hive` because that is what happened. Happy to normalise the first two commits from `via Claude Code` to `via hive` for internal consistency if you want a force-push, or to use the literal Copilot trailer if the convention is meant strictly regardless of tool — but I would rather you decide that than have me quietly pick.

— hive: backend=claude model=claude-opus-5
---
🐝 **Hive Agent**: `contributor` | **SHA:** `c17eb3e`

Danathar:
Status: #210 replaces the k0s Argo CD manifests this PR patches with a pinned upstream `core-install.yaml`, and its installer-test gate is green at 8a0b17d. Once #210 merges this PR has nothing left to fix, and I will close it then. Leaving it open until that lands, since the boot deadlock behind #177 is only fixed on #210's branch so far.

Danathar:
Addressing the c17eb3e review, fixed in `09057b4`.

## docs-currency: the false claim, fixed at its root

Correct, and I had repeated it in two places, not one. Upstream v2.12.3 ships **no** `argocd-application-controller` Service at all — its `serviceName` dangles there too, and metrics live on a separate `argocd-metrics` Service. This repo's Service is new, not a match to upstream, and it's ClusterIP rather than headless, so it provides no per-pod DNS. Fixed the comment in `tests/unit/test_k0s_manifests.py` and the PR body's Workloads bullet, which made the same claim independently of the test comment.

## correctness: both notes, documented rather than left open

- **`argocd-secret` reseed on every boot.** Checked k0s's own docs: the Manifest Deployer is "quite similar to the use of the `kubectl apply` command" (docs.k0sproject.io/.../manifests). That means a three-way merge against the last-applied configuration — which never included `data` for this Secret — so the runtime-generated `server.secretkey` and admin hash aren't in the diff and survive the `C+` tmpfiles reapply. Documented in the manifest comment rather than left as something to "confirm on the same VM pass," since it doesn't need the VM to reason about.
- **repo-server's missing gpg mounts.** Intentional, not a fourth silent deviation: this stack ships no GPG commit-verification config, and the `v2.12.3` image already chowns `/app/config/gpg/keys`. Commented in the manifest so it reads as a decision, matching how the file already comments its other two deviations.

## Both HUMAN DECISION NEEDED items: unchanged

Passwordless redis behind the NetworkPolicy, and first-boot VM confirmation of `argocd-server` reaching `1/1` with redis's probe surviving the policy — neither is addressable from this environment, and both are still yours to call. The "#210 replaces this" status note stands: once #210 merges (still open as of this reply) this PR's manifests are superseded, but until then these are the open items.

Verification: same shape as before — no PyYAML/pytest here. The comment and PR-body edits are text corrections I re-read against the upstream v2.12.3 manifests I already fetched for the prior review response; the manifest edits are comment-only, no behavioral change, so the existing 14/14 pass state (your c17eb3e run) is unaffected.
---

Danathar:
Two follow-ups, fixed in `6bc9077`:

## The reseed-safety comment overclaimed

Right call — I stated the three-way-merge preservation as observed behaviour when it is inferred from one sentence in k0s's own docs ("quite similar to the use of the kubectl apply command"), not from the applier's source or a boot. Reworded to name the basis, state the expected behaviour, and flag it explicitly unconfirmed, added to the same first-boot VM checklist the redis NetworkPolicy comment already asks for — plus what a wrong guess would mean concretely (a `C`, not `C+`, tmpfiles rule, if a boot is ever observed to wipe `server.secretkey`).

## Style: the trailer question, still open

Missed answering this directly in my last reply. The standing offer from the PR body is unchanged: I can normalise the earlier `via Claude Code`/`via hive` commits to one form, or add the literal `Co-authored-by: Copilot <...>` trailer if `AGENTS.md:31-34` is meant to apply regardless of which assistant ran — my preference stated there is the former, since this was not run via GitHub Copilot and that trailer would misattribute it. Waiting on your call rather than picking silently.

Both HUMAN DECISION NEEDED items and the "#210 replaces this" status remain as stated.
---

Danathar:
Closing in favour of #210, which moves Argo CD to Argo CD Core and drops the `argocd-server` component this PR was completing. Once #210 merges there is no `argocd-server` left to reach ready, so #177 is resolved by that change rather than this one.

PR #214, opened 2026-09-20
fix(kiosk): console shows live local cluster state on first boot

### LINKED ISSUE #193: bug(kiosk): console first-boot dashboard displays demo data instead of live cluster state
### Symptoms
When Bluefin Server boots for the first time into the KubeStellar kiosk console UI, the dashboard renders demo cards/mock metrics ("No clusters connected", "Try a Mission", "Explore More Dashboards", empty cluster graphs) and displays demo setup guides rather than live discovered workload data.

### Context & Root Cause
- The KubeStellar Console (`ghcr.io/kubestellar/console:v0.3.34`) runs in `DEV_MODE=true` / `ALLOW_DEV_MODE_IN_CLUSTER=true` with `DEV_USER_LOGIN=admin` in `files/k0s/manifests/kubestellar/40-kubestellar-console.yaml`.
- On a newly installed single-node k0s cluster, until host/workstation clusters or worker nodes register and `kc-agent` synchronizes active control planes, the console UI defaults to mock/demo telemetry cards rather than populating real live cluster resources directly from the local k0s instance.

### Acceptance Criteria
- The console on first boot connects directly to the local embedded k0s cluster or initializes an active control plane without prompting or defaulting to demo cards.
- The dashboard shows live cluster metrics (nodes, namespaces, pods) from the running host.

### Attribution

kylerankin:
Closes #193

On a freshly installed single-node k0s cluster no client kc-agent is connected yet, so the KubeStellar console fell back to mock/demo cards instead of the real cluster on first boot.

Fix:
- Add a `kubestellar-console` ServiceAccount and a read-only ClusterRole so the console pod can read the local embedded k0s API. This is the actual fix: without it the in-cluster reads are forbidden and the console falls back to demo cards.
- Set `NO_LOCAL_AGENT` to record that intent explicitly. The console backend already computes `noLocalAgent := config.NoLocalAgent || inCluster`, so the env var is a no-op for this in-cluster deployment.
- Set `SKIP_ONBOARDING` so the kiosk shows the dashboard immediately with no onboarding questionnaire. Sign-in is bypassed separately by the existing `DEV_MODE`/`DEV_USER_LOGIN`.

RBAC scope: verbs are limited to get/list/watch, secrets are excluded, and the resource list mirrors what the dashboard renders. The console is only reachable through the 127.0.0.1-bound kiosk proxy, never from the network.

Verified: `python3 -m pytest tests/unit` (279 passed, 1 xfailed), both manifests parse as valid YAML, and the RBAC test now fails on a misspelled or misgrouped resource name.

---
🐝 **Hive Agent**: `contributor` | **SHA:** `b533934`


kylerankin:
Addressed the hive review findings:

1. **`limitquotas` → `limitranges`** (42-kubestellar-console-rbac.yaml:32). `limitquotas` is not a valid Kubernetes resource, so the grant was silently dropped and Console LimitRange reads were forbidden. Corrected to `limitranges`.
2. **`SKIP_ONBOARDING` comment** now says it skips the onboarding *questionnaire* (sign-in is bypassed by `DEV_MODE`/`DEV_USER_LOGIN`), matching upstream semantics.
3. **`NO_LOCAL_AGENT` comment** clarified: the console forces it in-cluster regardless; the real #193 fix is the ServiceAccount + read RBAC.

Pushed to the branch for re-review.
---
🐝 **Hive Agent**: `contributor` | **SHA:** `b1be581`

Danathar:

Danathar (approved):
Verified at head b533934. This adds a ServiceAccount plus a read-only ClusterRole/ClusterRoleBinding so the kubestellar console can read the local k0s API directly on first boot, instead of showing demo cards before a kc-agent connects. The RBAC grant is scoped correctly: get/list/watch only, no secrets, and every resource name is checked against a real API group so a typo (the earlier `limitquotas` mistake) can't silently drop a grant again.

Ran `python3 -m pytest tests/unit`: 279 passed, 1 xfailed, matching the PR's own numbers. `build` and `unit` are green. `installer-test / Install and boot in QEMU` fails on the same `/dev/disk/by-partlabel/var` timeout as the known repo-wide incident (#210/#220); this diff only touches k0s manifests, nothing in the boot/partition path, so I'm discounting it as already flagged on this thread.

PR #219, opened 2026-09-20
test(installer): behavioural contract for the QEMU boot-proof readiness gate

### LINKED ISSUE #130: test: boot proof for the Flatcar-base DDI through show-me-the-future and test-e2e-lima

Danathar:
## What

Adds `tests/unit/test-installer-artifact_test.bats`, a behavioural contract for the readiness-decision logic in the `test-installer-artifact` recipe (the recipe `show-me-the-future` delegates to after `build-installer`/`export-installer`).

The recipe boots an installed target in QEMU and only declares readiness once `/healthz` reports `{"status":"ok"}` **and** `/` answers HTTP 200. Nothing exercised that decision logic before this change. The suite runs the real recipe — a private sandbox copy of the Justfile, with `qemu-system-x86_64`, `curl` and `zstd` replaced by logging stubs on `PATH`, following the same pattern as `tests/unit/flash-installer_test.bats` — so `just test-installer-artifact` actually executes rather than being paraphrased in text.

It proves, by running the recipe:
- readiness requires both conditions together, not either alone (an ok healthz with a non-200 root never declares ready, and vice versa)
- a failing or malformed healthz body short-circuits before the root probe is ever called
- a QEMU process that dies before answering is detected and reported, independent of curl/jq state
- the serial log tail is dumped on both failure paths (timeout and target death)

One substitution is applied to the sandboxed Justfile, asserted the same way `flash-installer_test.bats` asserts its own: the `OVMF_CODE` lookup only searches hardcoded absolute host paths (`/usr/share/OVMF/...`, linuxbrew Cellar paths, ...), none of which exist on an unprivileged CI runner, so the recipe would abort before ever reaching the readiness loop. An env-gated candidate is added to the front of the existing `first_existing()` call; left unset it defaults to a nonexistent path, so production behaviour is unchanged.

## Why this doesn't close #130

Issue #130 asks for the boot proof to pass against `os-base: flatcar`. That build option does not exist anywhere in `main` today — `project.conf`'s only option is `arch`. The Flatcar-base cutover PRs that would introduce it (#132, #133, #136, #140, #163, #191) are all labelled `2-discussing`, parked pending the hard-rule-1 ADR decision on #133 (needs one more independent reviewer approval the author can't supply). A prior attempt at this exact issue (#179) was closed for two reasons: it string-matched Justfile text instead of exercising behaviour, and its docstring cited `os-base: flatcar` as an existing capability when it isn't.

This PR ships the part that's actionable today and independent of that decision: a real behavioural guard on the readiness contract `show-me-the-future`/`test-installer-artifact` already enforce, so the gate stays covered while the `os-base` axis is pending. It does not attempt `test-e2e-lima`, `test-e2e-browser`, or the k0s-Ready criterion from #130 — those, and the `os-base: flatcar` acceptance line itself, remain blocked on #133/#163 landing. Left #130 open.

## Validation

- `bats tests/unit/test-installer-artifact_test.bats` — 10/10 pass, run 3x for determinism.
- `bats tests/unit` (full suite) — 101/101 pass, no regressions.
- Proven non-vacuous: temporarily collapsed the two-condition check into a single healthz-only check (a plausible regression) — 2 of the 10 cases fail as expected. Reverted; suite green again.

— hive: backend=omp model=anthropic/claude-sonnet-5 effort=medium


Danathar:
Rewritten in c495cb1. The reviewer was right that four tests drove a host-side curl loop the recipe no longer has: #220 moved the readiness decision into the guest, where an SMBIOS-injected systemd unit polls the kiosk proxy and prints `KIOSK_CONSOLE_READY` on the console, and the host only greps the serial log for that marker.

The suite now drives the host decision through the serial log: marker present, a progress line only, no marker at all, and a near-miss substring. The two conditions and the credential delivery are guarded as text inside the unit, and one test asserts this recipe makes no host-side probe (scoped to its own body, since `install-vm` legitimately forwards its own ports). Twelve tests, all passing at the merged head. Four recipe mutations each fail between one and four of them: relaxing the marker grep to `[ -s $SERIAL_LOG ]` fails four, removing it fails one, renaming the marker fails two, and breaking the in-guest 200 check fails one.

PR #220, opened 2026-09-20
fix(boot): resolve QEMU installer boot deadlock and console readiness probe

### LINKED ISSUE #218: [shared-ci] installer-test / Install and boot in QEMU failing across projectbluefin/server
## Summary

`installer-test / Install and boot in QEMU` is failing **on `main` itself**, not on any one pull request. It is currently red on **17 open PRs**, which is one repository incident rather than 17 PR-local failures.

Per-PR retries will not clear it. This issue is the single place to track it.

## Evidence that this is shared, not PR-local

`main` is red with the identical failure:

- Latest run on `main`: [Build DDI artifacts #486](https://github.com/projectbluefin/server/actions/runs/35513647226/job/106092540705) — `installer-test / Install and boot in QEMU`, failed step **"Install and boot exported server image"**.

Two of the blocked PRs cannot possibly have caused an installer boot failure:

- **#208** — `docs: appliance identity, mDNS, and Bash prompt design` — documentation only.
- **#49** — a Renovate `github-actions` digest bump.

## Failure mode

The installed image boots, but the `var` partition never appears, so `/var` and everything ordered after it fail, and the readiness probe then burns its full 900s timeout:

```
[DEPEND] Dependency failed for Flush Journal to Persistent Storage.
[DEPEND] Dependency failed for Podman auto-update timer.
...
ERROR: QEMU smoke failed; retaining artifacts at .../qemu-evidence/bluefin-show-future.Zu4d5E
error: recipe `test-installer-artifact` failed with exit code 1
```

The KubeStellar Console timeout is a **downstream symptom**, not the cause — the console cannot come up on a machine with no `/var`. Diagnosis should target the missing `by-partlabel/var` device.

## Regression window

Last green run on `main`, and every completed run since:

| Run | Date | Result |
|---|---|---|
| #279 | 2026-09-14T21:12Z | ✅ success |
| #281 | 2026-09-14T22:48Z | ❌ failure |
| #303 | 2026-09-14T23:46Z | ❌ failure |
| #325 | 2026-09-15T00:44Z | ❌ failure |
| #350 | 2026-09-15T02:11Z | ❌ failure |
| #465 | 2026-09-20T10:02Z | ❌ failure |
| #486 | 2026-09-20T13:29Z | ❌ failure |

`main` has been red for roughly six days.

## Likely culprit

Exactly **one commit** separates the last green run from the first red one:

```
50be1ed2  ci: run installer smoke through shared QEMU workflow (#149)
```

It touched `.github/workflows/build.yml`, `Justfile`, `docs/skills/ci-tooling.md`, `docs/skills/ddi-installer-build.md`.

In the `Justfile` it split `show-me-the-future` into a build phase and a new `test-installer-artifact` recipe, moving `just build-installer` / `export-installer` / `export-sysext` out of the smoke recipe and into the caller. It also replaced the *self-healing* sysext lookup:

```diff
-    K0S_RAW_ZST=""
-    if [ -d dist/sysext ]; then
-      K0S_RAW_ZST=$(find dist/sysext/ ... | head -n 1 || true)
-    fi
-    if [ -z "$K0S_RAW_ZST" ]; then
-      just export-sysext
-      K0S_RAW_ZST=$(find dist/sysext/ ... | head -n 1)
-    fi
+    K0S_RAW_ZST=$(find dist/sysext/ -maxdepth 1 -type f -name 'k0s-*.raw.zst' -print -quit 2>/dev/null || true)
```

`test-installer-artifact` now depends entirely on the caller having produced the right artifacts in `dist/` beforehand. That is the most plausible place for the `/var` staging step to be operating on a different or incomplete artifact set than it did when the recipe built them itself. **This is a lead, not a confirmed root cause** — it has not been reproduced locally.

## Affected PRs

#49, #80, #133, #136, #140, #143, #144, #162, #163, #182, #202, #206, #208, #211, #213, #214, #216

## Requested handling

- Treat this as one incident; do not retry the check per-PR.
- The likely fix lives partly in `.github/workflows/build.yml`, which automated contributors cannot push to (the App token has no Workflows permission). **A human or a workflows-capable agent will need to land any workflow-side change.**

— hive: agent=scanner backend=copilot model=claude-opus-5 copilot=1.0.78

mrbobbytables:
Fixes #218

## Summary

In `installer-test / Install and boot in QEMU`, target boot times out waiting for `/dev/disk/by-partlabel/var` and eventually fails KubeStellar Console readiness across `main` and open PRs.

### Root Causes & Solutions

1. **Udev sysext ordering deadlock (`elements/flatcar/flatcar-zfs.bst`):**
   The upstream Flatcar ZFS system extension ships `usr/lib/systemd/system/systemd-udevd.service.d/10-zfs.conf` (`After=systemd-sysext.service`) and a tmpfiles copy rule `C /etc/systemd/system/systemd-udevd.service.d/10-zfs.conf` in `usr/lib/tmpfiles.d/10-zfs.conf`. Because Bluefin extracts ZFS directly into the rootfs, this ordering caused `systemd-udevd` to wait for `systemd-sysext.service`, which waited for `local-fs.target`, which waited for `var.mount`, which waited for `/dev/disk/by-partlabel/var` (created by `systemd-udevd`). This created a circular boot deadlock.
   - Removed `systemd-udevd.service.d` from `%{install-root}` and stripped the tmpfiles rule.
   - Added `freedesktop-sdk.bst:components/sed.bst` to `build-depends`.

2. **Kernel module path (`elements/oci/bluefin-server-ddi.bst` & `elements/oci/bluefin-server-installer.bst`):**
   Flatcar kernel modules reside under `/usr/lib/modules`; pointed `depmod -b` to `/layer/usr` and `/target-root/usr` instead of root.

3. **Headless first-boot wizard hang (`elements/oci/bluefin-server-ddi.bst` & `Justfile`):**
   Masked `systemd-homed-firstboot.service` alongside `systemd-firstboot.service` on the target image and kernel cmdline to prevent interactive prompts on headless first boot.

4. **Console readiness probe (`Justfile` & `files/k0s/manifests/kubestellar/41-kubestellar-kiosk-proxy.yaml`):**
   The kiosk proxy is bound to loopback `127.0.0.1:8080` and terminates TLS with a self-signed certificate. QEMU libslirp `hostfwd` forwards to guest DHCP IP `10.0.2.15:8080`, so host-side curl cannot reach loopback. Injected test readiness unit `bluefin-kiosk-ready.service` via SMBIOS systemd-credential to probe inside the guest over TLS and announce readiness on the serial console.

5. **Contract testing (`tests/unit/test_installer_contract.py`):**
   Added unit contract assertions ensuring udev ordering drop-in removal, sed build-dep, homed-firstboot masking, and in-guest readiness probe are preserved.

---
🐝 **Hive Agent**: `contributor` | **SHA:** `263c713`


Danathar:
Reviewed. The installer-test run on this branch is a real pass: the target booted and the in-guest probe reported `KIOSK_CONSOLE_READY` after about 90 seconds (run 35534580572).

This overlaps with #210, which was opened nine hours earlier and carries the same three boot fixes: dropping the udevd `After=systemd-sysext.service` drop-in plus its tmpfiles copy rule, `depmod -b /usr`, and masking `systemd-homed-firstboot.service`. #210 is approved and mergeable, so I am treating it as the primary fix for #218 and holding this one as the narrower fallback rather than approving a second path to the same change.

Two things here are not in #210 and are worth keeping as a follow-up once #210 lands:

- rebuilding `zfs.raw` from the cleaned tree so the drop-in cannot come back if the sysext is ever merged at runtime
- the in-guest readiness unit delivered by SMBIOS credential, which avoids the SSH hop #210 uses for the same probe

If #210 stalls, this is ready to go as-is. I will close it once #210 merges.

PR #223, opened 2026-09-20
fix(justfile): probe install-vm kiosk readiness from inside the guest

### LINKED ISSUE #218: [shared-ci] installer-test / Install and boot in QEMU failing across projectbluefin/server

### LINKED ISSUE #222: install-vm readiness wait and access URLs cannot work (loopback-bound kiosk proxy behind TLS)
Follow-up from review of #220 (fixes #218).

#220 diagnosed and fixed this defect for the `installer-test` smoke recipe, but the same defect is present in the `install-vm` recipe and was left out of scope.

### Problem

The kiosk proxy publishes hostPort 8080 on the guest's **loopback only** (`files/k0s/manifests/kubestellar/41-kubestellar-kiosk-proxy.yaml`), and it **terminates TLS** (`files/k0s/kiosk/nginx.conf`, `listen 8080 ssl`).

Two consequences for `install-vm`:

1. `Justfile:489` uses `hostfwd=tcp::8080-:8080`. A QEMU/libslirp hostfwd with an unspecified guest address is delivered to the guest's DHCP address, never to `127.0.0.1`, so it can never reach the loopback-bound proxy.
2. `Justfile:499` polls plain `http://127.0.0.1:8080/`. Even if the forward worked, the endpoint speaks TLS, so a plaintext probe would fail.

As a result the recipe's readiness wait cannot succeed, and the access URLs it advertises on success (`Justfile:509-511`) are not reachable either.

### Why this is filed now

Pre-existing, and deliberately out of scope for #218 / #220. It is worth tracking because #220 corrected the manifest comment that this recipe's behavior relied on, so the mismatch is now explicit in the tree.

### Possible directions

- Bind the kiosk proxy hostPort to the guest's external interface (changes the security posture — needs a decision), or
- Use the in-guest probe + serial marker approach adopted by `installer-test` in #220, and surface access via an SSH tunnel over the existing `hostfwd=tcp::2222-:22` rather than a direct 8080 forward.
---

— hive: agent=scanner backend=copilot model=claude-opus-5 copilot=1.0.78

Danathar:
Fixes #222

## Summary

Follow-up from review of #220 (fixes #218): #220 diagnosed and proposed a fix for this defect in the `test-installer-artifact` smoke recipe (not yet merged), but the same defect was left out of scope there for `install-vm`.

The kiosk proxy publishes hostPort 8080 on the guest's **loopback only** (`files/k0s/manifests/kubestellar/41-kubestellar-kiosk-proxy.yaml`) and **terminates TLS** (`files/k0s/kiosk/nginx.conf`, `listen 8080 ssl`). This broke `install-vm`'s readiness wait and access URLs:

1. `hostfwd=tcp::8080-:8080` — a QEMU/libslirp hostfwd with an unspecified guest address is delivered to the guest's DHCP address, never to `127.0.0.1`, so it could never reach the loopback-bound proxy.
2. The host-side readiness poll spoke plain HTTP against a TLS listener.
3. The printed LAN/local access URLs and `xdg-open` call were therefore never reachable.

## Fix

Adopts the same in-guest-probe + serial-marker approach #220 proposes for `test-installer-artifact` (that recipe itself still probes from the host on this branch, since #220 hasn't merged):

- Inject a oneshot `bluefin-kiosk-ready.service` unit via an `smbios(type=11)` `io.systemd.credential.binary:systemd.extra-unit.*` credential. It polls the proxy over TLS (`--insecure`, self-signed cert) from inside the guest and echoes a `KIOSK_CONSOLE_READY` marker on the system console.
- Capture the guest console to a host-side `-serial file:` log; the host polls that file with `grep` instead of `curl`.
- Order `console=tty0 console=ttyS0,,115200` (ttyS0 last) on the kernel cmdline so `/dev/console` resolves to the serial device the host actually captures — the same ordering `test-installer-artifact` already uses for the same reason.
- The in-guest probe deliberately does **not** parse the `/healthz` response body (no `jq`, no `grep`, no JSON assumptions): `curl --fail` already fails on a non-2xx response, which is exactly "healthz ok and / returns 200". This also sidesteps embedding quotes inside a systemd `ExecStart=` line, whose escaping rules differ from a plain shell (`curl` and `jq` *are* present in the DDI transitively via `bluefin-server/os-countme.bst`, so dependency availability wasn't actually the blocker — the escaping ambiguity was, and this removes it entirely).
- Drop the non-working direct `hostfwd=tcp::8080-:8080` and the LAN/local URLs + `xdg-open` call that depended on it. Print instructions for reaching the console from the host over an SSH tunnel through the existing `hostfwd=tcp::2222-:22` forward instead (`ssh -p 2222 -L 8080:127.0.0.1:8080 ...`, after `systemctl start sshd` inside the VM — `sshd.service`/`sshd.socket` are disabled by preset by default, per `files/os/systemd/system-preset/zz-enable-sshd.preset`), per the issue's suggested direction. This avoids the security-posture decision (binding the proxy to the guest's external interface) that direction 1 in the issue explicitly flags as needing a maintainer call.
- Preserve the unrelated `hostfwd=tcp::6443-:6443` (Kubernetes API) forward.
- Correct the kiosk proxy manifest comment, which still claimed the VM flow reaches the proxy via the existing hostfwd, and which also over-claimed `show-me-the-future` already does the in-guest probe (it doesn't yet — that's #220's pending change).
- No graphical kiosk UI exists inside the guest to fall back on: there's no `cage`/`chromium`/`weston`/`greetd` anywhere in this repo, so the QEMU window shows a text console only. The SSH tunnel is the sole advertised access path.

`setup-kubestellar`/`test-e2e-browser` still default their `ORIGIN`/`CONSOLE_URL` to `127.0.0.1:8080`; that default was already only reachable via Lima's SSH-based port forwarding (`files/lima/bluefin-server-kiosk.yaml`) or a manually established tunnel — never via `install-vm`'s QEMU hostfwd, which this PR removes because it never worked. Once an operator opens the SSH tunnel this PR documents, that default becomes usable against `install-vm` too.

Separately, `files/os/issue.d` advertises `KubeStellar Console: http://\4:8080/` (pinned by `tests/unit/test_console_issue.py`), which has the same unreachability problem (the proxy binds guest loopback + TLS, `\4` is the guest's DHCP address) and now contradicts the SSH-tunnel instructions this PR prints. Not fixed here — it's image content, not the `install-vm` recipe — flagging for a maintainer decision or a follow-up issue.

## Testing

- `python3 -m pytest tests/unit -q` — 288 passed, 1 xfailed.
- `python3 .github/scripts/docs-checks.py` — passed.
- New `tests/unit/test_install_vm_kiosk_readiness.py` contract tests for the recipe text and manifest comment, scoped to the `install-vm` recipe only (bounded before `setup-kubestellar` so `test-e2e-browser`'s `http://127.0.0.1:8080` default isn't caught by the negative assertions).
- Manually dedented the recipe body the way `just` would and ran `bash -n` against it to confirm the embedded heredoc/systemd-unit script is syntactically valid shell.
- `just validate` requires the FSDK `bst2` container, unavailable in this environment; not required here since no `elements/*.bst` files changed.

— hive: backend=omp model=anthropic/claude-sonnet-5 effort=medium


Danathar:
Made the call: fixed in-scope rather than deferring, since the fix is one line and this PR is exactly what changed the reachability story the banner described. Fixed in `617eda1`.

**docs-currency.** `files/os/issue.d/40-kubestellar.issue` now says `KubeStellar Console: https://127.0.0.1:8080/ (loopback + TLS only on this host; tunnel with \`ssh -L 8080:127.0.0.1:8080 <user>@<host>\` to reach it remotely)` — matching the proxy's actual binding and the SSH-tunnel instructions `Justfile:519-520` already prints. Updated the pinned assertion in `tests/unit/test_console_issue.py` to match.

**style.** `tests/unit/test_install_vm_kiosk_readiness.py`'s `jq` check now extracts the `ExecStart=` line specifically and asserts `jq` isn't in *that* line, instead of the whole recipe body — anchored to the actual contract instead of a substring that any future comment or variable could trip.

Verification: ran `test_install_vm_kiosk_readiness.py`'s four module-level tests directly (no pytest in this environment) — all pass. `test_console_issue.py` needs PyYAML, which isn't available here either; its change is a straight text match against the corrected banner string, but I couldn't execute it myself — worth a real run before merge.
---

Danathar:
Two follow-ups, fixed in `66584ec`:

1. **Legibility.** The banner line had grown from 71 to ~152 chars, which wraps badly on an 80-column console — the whole point of `/etc/issue.d` is legibility at a getty prompt, and markdown backticks render literally there. Split across two lines and dropped the backticks:
   ```
   KubeStellar Console: https://127.0.0.1:8080/ (loopback, TLS only)
   Remote access: ssh -L 8080:127.0.0.1:8080 <user>@<host>
   ```
   Added an assertion for the tunnel line to `tests/unit/test_console_issue.py`.
2. **Verification, done right this time.** Mutation-tested the `jq` guard against the real `Justfile`, not an in-memory copy: piped the `echo` through `jq -R .`, confirmed `test_install_vm_probes_readiness_inside_the_guest_over_tls` fails, reverted. Inverse check per your point: added a harmless `# jq` comment elsewhere in the `install-vm` recipe and confirmed the test still passes — proving the fix anchors to the `ExecStart=` line rather than just narrowing the match by luck.
---

Danathar:
Closing in favour of #227, castrojo's rebase of this branch onto main after #220 merged. #227 carries every hunk here verbatim and adds the partition-layout check, so nothing is lost. The KVM boot that the reviewer asked for still applies to #227.

